PR #6, opened 2018-02-14
Gel panel mapping

gkos-bio:
refactor code to access Google Cloud Bucket
refactor GenomicsEngland Panel App to map old panel id to new panel id

### DIFF
--- a/common/GCSUtils.py
+++ b/common/GCSUtils.py
@@ -1,0 +1,35 @@
+from settings import Config
+import logging
+import google.cloud.storage as gcs
+import google.cloud.exceptions as gce
+import tempfile
+import gzip
+
+
+class GCSBucketManager(object):
+
+    def __init__(self):
+        print("init GCSBucketManager")
+        super(GCSBucketManager, self).__init__()
+        self._logger = logging.getLogger(__name__)
+        self.gcs_client = gcs.Client(Config.GOOGLE_DEFAULT_PROJECT)
+        self.bucket = None
+        self._logger.info("This is done")
+
+        try:
+            for bucket in self.gcs_client.list_buckets():
+                print(bucket)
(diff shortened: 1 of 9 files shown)


common/GCSUtils.py
```diff
@@ -0,0 +1,35 @@
...
+        self._logger = logging.getLogger(__name__)
+        self.gcs_client = gcs.Client(Config.GOOGLE_DEFAULT_PROJECT)
+        self.bucket = None
+        self._logger.info("This is done")
+
+        try:
+            for bucket in self.gcs_client.list_buckets():
+                print(bucket)
+            self.bucket = self.gcs_client.get_bucket(Config.GOOGLE_BUCKET_EVIDENCE_INPUT)
+        except gce.NotFound:
```
elipapa: i think we should avoid using credentials _for downloads_ if we can.

modules/Gene2Phenotype.py
```diff
@@ -60,12 +54,11 @@ def process_g2p(self):
...
 
     def generate_evidence_strings(self, filename):
+
         total_efo = 0
-        blob = self.bucket.get_blob(filename)
-        with open('/tmp/%s'%(G2P_FILENAME), 'wb') as file_obj:
-            blob.download_to_file(file_obj)
 
-        with gzip.open('/tmp/%s'%(G2P_FILENAME), mode='rt') as zf:
+        with gzip.open(self.get_gcs_filename(filename), mode='rt') as zf:
```
elipapa: it's unnecessary to call the google cloud storage if the link is public. 
instead of creating an helper function, you should use something like:
```python
with requests.get(url, stream=True) as r:
     for row in csv.DictReader(r.iter_lines(decode_unicode=True)):
          print(row)
```

PR #7, opened 2018-02-14
refactor to simplify logic and make each module independent (aka. make phewas step run again(

elipapa:
- i removed commandline.py. Each script should be called with `python modules/data/run.py` 
- each script can now import from `common` without having to use the commnadline imports logic
- removed setup.py

for phewas:
- separated mapping generation step and evidence creation step. That way we can have a reference mapping file that we update.


elipapa:
merging for the lack of reviews 😀

PR #11, opened 2018-04-16
Ck progeny

ckongEbi:


README.md
```diff
@@ -1,36 +1,4 @@
-# phewascatalog_parser
-script to convert phewascatalog database to JSONs, with ENSGIDs and EFO ids.
+# Main Wiki Page 
+* on how to setup and run the various evidence parsers
+* https://github.com/opentargets/data_release/wiki/Data-Generation-(Evidence)
```
elipapa: i think the documentation should live in this repo's README

modules/GEPanelApp.py
```diff
@@ -0,0 +1,248 @@
...
+from ontoma import OnToma
+from settings import Config
+from collections import OrderedDict
+from common.HGNCParser import GeneParser
+from common.GCSUtils import GCSBucketManager
+
+import logging
+logger = logging.getLogger(__name__)
+
+__copyright__ = "Copyright 2014-2018, Open Targets"
```
elipapa: i will forever be against including these `__uselessinformation__` strings in the code. In my mind this information is already contained in the LICENSE file, the README and in the git commit histories. Adding it here will only mean that it will soon become out of date.

what difference does it make if the `__status__` is not production?

modules/GEPanelApp.py
```diff
@@ -0,0 +1,248 @@
...
+        self.panel_app_info = list()
+        self.map_strings = OrderedDict()
+
+    def process_ge(self):
+
+        self.ontoma = OnToma()
+        gene_parser = GeneParser()
+        gene_parser._get_hgnc_data_from_json()
+
+        self.omim_to_efo_map = self.ontoma._omim_to_efo
```
elipapa: ontoma already implements a `omim_lookup` https://github.com/opentargets/OnToma/blob/master/ontoma/interface.py#L259. shouldn't we use this instead of accessing the _omim_to_efo private object ?

using the function allows us to change the logic in the future if needed.

modules/GEPanelApp.py
```diff
@@ -0,0 +1,248 @@
...
+        OMIM  130650
+        #120330:Papillorenal syndrome
+  ** If NO EFO mapping found try OntoMA
+    i.e Verheij syndrome,  615583
+        => Ontoma: http://www.ebi.ac.uk/efo/EFO_0000699,
+* If NO OMIM id found, try OntoMa
+    i.e. Diffuse keratoderma with knuckle pads
+        => Ontoma: 'http://www.orpha.net/ORDO/Orphanet_2698'
+
+* The OntoMa mapping process
```
elipapa: this belongs to ontoma's documentation and it risks becoming outdated here. 
I thought it was in the ontoma documentation but being in a private method it did not render.. thanks for pointing it out! 

Now a link to the ontoma docs should work... for eg. http://ontoma.readthedocs.io/en/stable/ontoma.html#ontoma.interface.OnToma.find_term

modules/GEPanelApp.py
```diff
@@ -0,0 +1,248 @@
...
+        self.ontoma = OnToma()
+        gene_parser = GeneParser()
+        gene_parser._get_hgnc_data_from_json()
+
+        self.omim_to_efo_map = self.ontoma._omim_to_efo
+        self.symbol_to_ensembl_map = gene_parser.genes
+
+        print("Parsed %s OMIM to EFO mapping " % len(self.omim_to_efo_map))
+        print("Parsed %s ZOOMA to EFO mapping " % len(self.ontoma._zooma_to_efo_map))
+        print("Parsed %s Name to EFO mapping " % len(self.ontoma.name_to_efo))
```
elipapa: i have added this log message to ontoma. you should be able to remove them from here.

modules/GEPanelApp.py
```diff
@@ -0,0 +1,248 @@
...
+    def process_ge(self):
+
+        self.ontoma = OnToma()
+        gene_parser = GeneParser()
+        gene_parser._get_hgnc_data_from_json()
+
+        self.omim_to_efo_map = self.ontoma._omim_to_efo
+        self.symbol_to_ensembl_map = gene_parser.genes
+
+        print("Parsed %s OMIM to EFO mapping " % len(self.omim_to_efo_map))
```
elipapa: ontoma's logger should already output this information. See here: https://github.com/opentargets/OnToma/blob/master/ontoma/downloaders.py#L31

Does it not work ?

modules/GEPanelApp.py
```diff
@@ -0,0 +1,248 @@
...
+        --- below this line we might not have a term in the platform ---
+        5. HP OBO lookup
+        6. OLS API HP lookup - exact match
+        7. OLS API EFO lookup - not exact
+        (8. ?Zooma medium)
+
+'''
+
+#class GEPanelApp(GCSBucketManager):
+class GEPanelApp():
```
elipapa: Is there any reason to wrap the script logic and all the functions in a class?
I really can't see the point. I don't anticipate one would be ever needing more than a single class instance of GEPanelApp.

All the functions could live separately and then be run as part of the `main()` function. State could be preserved within variable declared in `main()`. 

That would also have the benefit of being able to run the file as a standalone script, if needed.

modules/GEPanelApp.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def process_ge_panel(self):
+        '''
+         Create panel app info list and phenotype set
+         :return: Unique phenotype list
+        '''
+        print("Retrieving GE PanelApp data...")
+        url = 'https://panelapp.genomicsengland.co.uk/WebServices/search_genes/all/'
+        phenotype_list = []
+        c = 0
```
elipapa: hykes! short variable names can be confusing!

modules/GEPanelApp.py
```diff
@@ -0,0 +1,248 @@
...
+        d = 0
+        e = 0
+        f = 0
+
+        for panel_name, panel_id, panel_version, panel_diseasegroup, panel_diseasesubgroup in self.get_panel_list():
```
elipapa: why don't you declare a `Panel` class? it might allow you to iterate through `Panel` objects more succinctly and parse each object with code kept outside of the request handling logic.

modules/GEPanelApp.py
```diff
@@ -0,0 +1,248 @@
...
+            #c += 1
+            print("Reading panel : %s %s %s %s %s" % (panel_name, panel_id, panel_version, panel_diseasegroup, panel_diseasesubgroup))
+            r = requests.get(url, params={"panel_name": panel_name}, timeout=30)
+            results = r.json()
+
+            ''' process genes in a panel '''
```
elipapa: this would be a good fit for the `__init__` method of the `Panel` class.

modules/PROGENY.py
```diff
@@ -0,0 +1,288 @@
...
+import sys
+import logging
+import datetime
+import opentargets.model.core as opentargets
+import opentargets.model.bioentity as bioentity
+import opentargets.model.evidence.core as evidence_core
+import opentargets.model.evidence.linkout as evidence_linkout
+import opentargets.model.evidence.association_score as association_score
+
+__copyright__ = "Copyright 2014-2018, Open Targets"
```
elipapa: as above, I'd recommend getting rid of these `__soontobeobsoletedocumentation__` strings.

modules/PROGENY.py
```diff
@@ -0,0 +1,288 @@
+from settings import Config
+from common.HGNCParser import GeneParser
+import sys
+import logging
+import datetime
+import opentargets.model.core as opentargets
+import opentargets.model.bioentity as bioentity
+import opentargets.model.evidence.core as evidence_core
```
elipapa: i'd recommend using the `pjs` package https://github.com/cwacek/python-jsonschema-objects

modules/PROGENY.py
```diff
@@ -0,0 +1,288 @@
...
+PROGENY_SYMBOL_MAPPING = {
+    'NKFB1'   : 'NFKB1',
+    'MAPK2K1' : 'PRKMK1',
+    'PI3K(Class1)' : 'PIK3CA',
+    'VEGFR'   : 'KDR',
+    'BCL-W'   : 'BCL2L2',
+    'BCL-XL'  : 'BCL2L1'
+}
+
+class PROGENY():
```
elipapa: as above, I am not sure if you need to have a class that runs the code.
all of the below could be contained in a `main()` method.
You need classes only when you want to make multiple independent instances containing different states. For example `evidenceStrings`, as you do below.

modules/PROGENY.py
```diff
@@ -0,0 +1,288 @@
...
+            )
+        )
+        error = provenance_type.validate(logging)
+
+        if error > 0:
+            self.logger.error(provenance_type.to_JSON(indentation=4))
+            sys.exit(1)
+
+        with open(filename, 'r') as progeny_input:
+            n = 0
```
elipapa: `pylint` wouldn't allow you to commit single letter variables.. very perl :)

elipapa:
CI is failing because the `ontologyutils` package we import is failing. @gkos-bio is there something we could fix there?

ckongEbi:
@elipapa @gkos-bio @mkarmona  @MichaelaEBI this branch currently includes parsers that generate JSON for `intogen` , `progeny` , `genomics_england`, `slapenrich`, `gene2phenotyp` .   Data for `18.06` are generated for these sources.  

Feel free to re-factor for best coding practices.

afaulconbridge:
Because this branch is now quite out of date from master, closing this pull request.

PR #17, opened 2018-10-16
Af progeny ukbiobank

afaulconbridge:
Based on previous pull request opentargets/evidence_datasource_parsers#11 from ck_progeny branch, I've extracted the progeny and ukbiobank files of interest to be added into the master branch.

If we find we need additional things moving over later, we can come back and extract the rest. But to keep it manageable, I think we should do it piece by piece to start with.


modules/UKBiobank.py
```diff
@@ -0,0 +1,201 @@
...
+
+__copyright__ = "Copyright 2014-2018, Open Targets"
+__credits__   = ["ChuangKee Ong"]
+__license__   = "Apache 2.0"
+__version__   = "1.2.8"
+__maintainer__= "ChuangKee Ong"
+__email__     = ["data@opentargets.org"]
+__status__    = "Production"
+
+class UKBiobank():
```
mkarmona: as python 3 you dont need to put that just `class UKBiobank:` otherwhise python2 `class UKBiobank(object):` 
https://stackoverflow.com/questions/4906014/python-default-inheritance

modules/UKBiobank.py
```diff
@@ -0,0 +1,201 @@
...
+        with open(filename, 'w') as ukbiobank_output:
+            n = 0
+            for evidence_string in self.evidence_strings:
+                n += 1
+                self.logger.info(evidence_string.disease.id[0])
+                # get max_phase_for_all_diseases
+                error = evidence_string.validate(logging)
+
+                if error == 0:
+                    print(evidence_string.to_JSON(indentation=None))
```
mkarmona: please, logging from the beginning makes the code better and more readable and debuggable

modules/UKBiobank.py
```diff
@@ -0,0 +1,201 @@
...
+
+        now = datetime.datetime.now()
+
+        '''
+            build evidence.provenance_type object
+        '''
+        provenance_type = evidence_core.BaseProvenance_Type(
+            database=evidence_core.BaseDatabase(
+                id="UKBiobank",
+                version='2018.04',
```
mkarmona: hiding constants are dangerous. could you please put at module level or so to easily check and change when it is necessary? thanks

modules/PROGENY.py
```diff
@@ -0,0 +1,287 @@
...
+    def build_evidence(self, filename=Config.PROGENY_FILENAME):
+
+        now = datetime.datetime.now()
+        '''
+            build evidence.provenance_type object
+        '''
+        provenance_type = evidence_core.BaseProvenance_Type(
+            database=evidence_core.BaseDatabase(
+                id="PROGENY",
+                version='2018.04',
```
mkarmona: same here? twice repeated. that is not good

settings.py
```diff
@@ -56,6 +56,14 @@ class Config:
     # HGNC
     GENES_HGNC = 'http://ftp.ebi.ac.uk/pub/databases/genenames/new/json/hgnc_complete_set.json'
 
+    # PROGENY
+    PROGENY_FILENAME = file_or_resource('progeny_normalVStumor_opentargets.txt')
+    PROGENY_EVIDENCE_FILENAME = HOME_DIR + '/progeny-20-05-2018.json'
```
MichaelaEBI: I guess that using the google buckets for internal evidence generation will help us avoid using 'HOME_DIR'?

MichaelaEBI: On CKs branch, the PheWAS file settings were also in this file. Now they live in their own file in [phewascat/constants.py](https://github.com/opentargets/evidence_datasource_parsers/blob/master/modules/phewascat/constants.py). It would be good to be consistent for all internal evidence types.

modules/UKBiobank.py
mkarmona: what I want to say is `()` are unnecesary

settings.py
afaulconbridge: Yes, but we should address the phewas stuff in a separate pull request.

PR #19, opened 2018-11-29
Only generate evidence strings if p-value <1e-4 and deal with p-value==0

MichaelaEBI:
JSON evidence strings are only generate if the p-value is < 1e-10 and p-values of 0 are set to 1e-20.


MichaelaEBI:
The generated JSON file validates almost all strings. The gene symbol USP12 does not return an ensembl ID for some reason. This affects 5 evidence strings and I will look into that after the 18.12 release. Everything else validates fine though!

PR #20, opened 2018-12-05
Ag fix phewas mapping

AsierGonzalez:
make_effo_mappings.py was not working because a file object was passed to os.path.getsize() instead of a path.


modules/SystemsBiology.py
```diff
@@ -0,0 +1,164 @@
...
+                # pmid	method	score_type
+                (sysbio_pmid, sysbio_method, score_type) = tuple(line.rstrip().split('\t'))
+                SYSBIO_PMID_MAP[sysbio_pmid]={'method':sysbio_method, 'score_type':score_type}
+
+        with open(filename1, 'r') as sysbio_input:
+            n = 0
+            for line in sysbio_input:
+                n +=1
+                if n>1:
+                    '''
```
mkarmona: please, use a proper comment as `#` but not strings

modules/SystemsBiology.py
```diff
@@ -0,0 +1,164 @@
...
+                        ##TODO issue with append, take only last item of the gene
+                        self.evidence_strings.append(evidenceString)
+
+                    else:
+                        self.logger.error("%s is not found in Ensembl" % target_name)
+
+            self.logger.error("%s evidence parsed"%(n-1))
+            self.logger.error("%s evidence created"%len(self.evidence_strings))
+
+        sysbio_input.close()
```
mkarmona: @AsierGonzalez you dont need to close as with `with` context does that for you

modules/SystemsBiology.py
```diff
@@ -0,0 +1,164 @@
...
+                        evidenceString.evidence.evidence_codes = ["http://purl.obolibrary.org/obo/ECO_0000053"]
+                        evidenceString.evidence.provenance_type = provenance_type
+                        evidenceString.evidence.resource_score = resource_score
+
+
+                        error = evidenceString.validate(logging)
+
+                        #print(evidenceString.to_JSON(indentation=None))
+                        ##TODO issue with append, take only last item of the gene
+                        self.evidence_strings.append(evidenceString)
```
mkarmona: @AsierGonzalez this limit your process to a small number of evidences not greater than your local machine memory. A proper approach is using an each line process and saving each line as all lines are independent. have a look at the package [pyfunctional](https://pypi.org/project/PyFunctional/) to solve this problem.. Also, it is always good having also the package [more-itertools](https://pypi.org/project/more-itertools/)

modules/SystemsBiology.py
```diff
@@ -0,0 +1,164 @@
...
+from common.HGNCParser import GeneParser
+import sys
+import logging
+import datetime
+import opentargets.model.core as opentargets
+import opentargets.model.bioentity as bioentity
+import opentargets.model.evidence.core as evidence_core
+import opentargets.model.evidence.association_score as association_score
+
+__copyright__ = "Copyright 2014-2018, Open Targets"
```
mkarmona: this `__*__` has to disappear in favor of having a LICENSE file in the root folder and also adding to the README.md a [copyright like this](https://github.com/opentargets/data_pipeline/blob/master/README.md#copyright)

modules/SystemsBiology.py
mkarmona: please, use a proper comment as `#` but not strings

modules/SystemsBiology.py
mkarmona: @AsierGonzalez this limit your process to a small number of evidences not greater than your local machine memory. A proper approach is using an each line process and saving each line as all lines are independent. have a look at the package [pyfunctional](https://pypi.org/project/PyFunctional/) to solve this problem.. Also, it is always good having also the package [more-itertools](https://pypi.org/project/more-itertools/)

modules/SystemsBiology.py
MichaelaEBI: This line is present in all these modules, so it probably makes sense to tidy them all up in one go.

modules/SystemsBiology.py
MichaelaEBI: I know and whenever I see a comment that is a string I quickly change it to a proper comment. They have all been changed in the sysbio.py file.

modules/SystemsBiology.py
MichaelaEBI: Asier and I are looking forward to your masterclass for modules :-)

modules/SystemsBiology.py
MichaelaEBI: This has been fixed in my branch ms_sysbio_module_2

PR #24, opened 2019-02-01
MS SysBio module version 2

MichaelaEBI:
The Systems Biology evidence scores are now calculated based on the p-values or rank-based scores provided. A small update to the settings file allowed for the gene2phenotype module to run.


AsierGonzalez:
> @MichaelaEBI the file resources/SysBioEvidence_Jan2019.tsv seems to be identical to resources/sysbio_evidence-31-01-2019.tsv but it isn't used anywhere, is it?

And it seems to be the same as resources/sysbio_evidence_nov2018.tsv, which is no longer used either. I don't see any reason to keep it.

MichaelaEBI:
@AsierGonzalez I will let you know when I have moved these files to their newly created evidence buckets.

MichaelaEBI:
There are two minor changes in sysbio_evidence-31-01-2019.tsv which means that two more rows can be processed by the pipeline (gene name corrections). It probably makes sense to approve this pull request and then start re-writing all the modules to access data files from the Google Bucket.

AsierGonzalez:
Approving this PR now, any updates related to input files will go in a new one.

PR #29, opened 2019-04-24
Explain PhenoDigm branches in README

AsierGonzalez:


AsierGonzalez:
I guess master should work but for some reason there are two incompatible versions of PhenoDigm in master and the other branch. At least now there is some information about it in the README.

PR #39, opened 2020-02-13
genetics portal evidence

DSuveges:
A script to generate evidences based on GWAS associations sourced from the Genetics portal.

* parameters read from command line.
* Duplicated evidences are excluded.
* JSON schema used to build the evidence.
* Logging implemented.

### DIFF
--- a/modules/GeneticsPortal.py
+++ b/modules/GeneticsPortal.py
@@ -9,27 +9,16 @@ import gzip
 import requests
 import logging
 
-class genetics_portal_evidence_generator(object):
+# Importing settings:
+from settings import Config
+
+class genetics_portal_evidence_generator(Config):
     """
     Based on the provided schema this class generates evidence string for the Genetics portal
     derived associations.
+
+    Parameters inherited from Config class
     """
-
-    ##
-    ## The following values should be read from a config file....
-    ##
-
-    # Defining base URLs for links:
-    activity_base_url = 'http://identifiers.org/cttv.activity'
-    target_id_base_url = 'http://identifiers.org/ensembl'
-    target_type_base_url = 'http://identifiers.org/cttv.target'
-    literature_base_url = 'http://europepmc.org/abstract/MED'
-    genetics_portal_base_url = 'https://genetics.opentargets.org'
-    disease_base_url = 'http://www.ebi.ac.uk/efo'
-    consequence_base_url = 'http://purl.obolibrary.org/obo'
-
-    # Evidence code: computational inference:
-    evidence_code = 'http://purl.obolibrary.org/obo/ECO_0000362'
 
     # Source name:
     source_id = 'ot_genetics_portal'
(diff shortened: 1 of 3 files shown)


modules/GeneticsPortal.py
```diff
@@ -0,0 +1,534 @@
...
+        return return_data
+
+
+def initialize_evidence_generation(df):
+    """
+    As it is not possible for multiprocessing.Pool to copy over states of objects,
+    this function instantiates the evidence generator class and generates evidences.
+    """
+
+    global schemaFile
```
mkarmona: the use of global is quite dangerous. I would say this case is not justified enough to bring a global value into a function used to map a collection. In that case, I would suggest the usage of [partial functions](https://docs.python.org/2/library/functools.html#functools.partial). A quick example to ilustrate the rewrite is drafted below

```python
>>> import functools as fn
>>> def func2(df, schema):
...     # df work
...     print schema
... 
>>> func1 = fn.partial(func2, schema="asdf")
>>> func1(42)
asdf

```

modules/GeneticsPortal.py
```diff
@@ -0,0 +1,534 @@
...
+def parallelize_dataframe(df, schema_url, n_cores=2):
+
+    # Splitting dataframe as many chunks as many cores we have defined:
+    df_split = np.array_split(df, n_cores)
+
+    # Initialize pool object:
+    pool = Pool(n_cores)
+
+    # Execute funciton on dataframe chunks and pool output:
+    df = pd.concat(pool.map(initialize_evidence_generation, df_split))
```
mkarmona: Why do you need to join back the files? Can you save each df chunk straight to a file? use a [uuid 4](https://stackoverflow.com/a/534847/1197021) to append to each filename with a prefix? that would help the provision of future scalability improvements.

modules/GeneticsPortal.py
```diff
@@ -0,0 +1,534 @@
...
+
+    # Closing pool object:
+    pool.close()
+    pool.join()
+
+    return df
+
+
+def remove_duplicates(df):
```
mkarmona: not sure why you need to detect duplicates. The pipeline process has to take care of it, either the current one or the rewrite.

DSuveges: One of @d0choa 's recent commit  ([ee503a7bd735201d9a5cc5fc54839671f8b98657](https://github.com/opentargets/evidence_datasource_parsers/commit/ee503a7bd735201d9a5cc5fc54839671f8b98657)) lists uniqueness as a requirements. But I agree, that is not critical for sure.

modules/GeneticsPortal.py
DSuveges: As suggested, a partial function is used, so no global variable is used in the code. Thanks for the suggestion.

modules/GeneticsPortal.py
```diff
@@ -0,0 +1,521 @@
...
+        self.reported_trait = row['trait_reported']
+        self.efo_id = row['efo']
+
+        # Locus to gene score:
+        self.l2g_score = row['y_proba_full_model']
+
+        ##
+        ## Applying a few checks before proceeding:
+        ##
+
```
DSuveges: The following condition is used to filter out evidences where disease ID is not available. Related to issue #858

PR #43, opened 2020-03-19
Update documentation with genetics portal evidence generation.

DSuveges:
The `README.md` file is extended with instructions on how the Genetics Portal evidences are generated.


AsierGonzalez:
Merging this PR but I suggest that the README is updated in the future to clarify what the values of some (e.g. `--cluster`, `--projects` and `--region`) of the parameters should be. Alternatively, as this parameters depend on the infrastructure used this information may need to be documented elsewhere.

PR #48, opened 2020-08-19
Ag g2p add all panels and more info

AsierGonzalez:
Two big changes:
- Added processing of Eye, Skin and Cancer gene panels
- Map disease names using OnToma instead of crossrefs of OMIM ids


AsierGonzalez:
Added handling of missing dates, only when available will `date_asserted` be set

DSuveges (approved):
All looks good to me.

PR #50, opened 2020-11-10
il-phewas-update 20/11

ireneisdoomed:
PR to update several variant-related fields with information coming from the Variant Index from Genetics Portal.

- `variant field `updated with:
    - "`id`" field, if present, with information identifying the variant: "`chromosome_position_reference allele_alternative_allele`", i.e. "4_109737911_T_C"";
    - `"rs_id"` field, with the corresponding rsid of the variant;

- `gene2variant` field updated with:
    - `"functional_consequence" `field is no longer hardcoded, but populated from the Variant Index data
    - Important to note: there are 903 070 null values for this field now in the input file. For these cases and given it is a required field, the hardcoded value is reset (`http://purl.obolibrary.org/obo/SO_0001060`).

- `unique_association_fields` updated with:
    - `"variant_id"` is now populated with the `"id"` field (see above), if present, and if not with the `"rs_id"` field

The script now checks if the evidence string is already resent in the output file, so that we can avoid duplicates. The implemented solution is extremely inefficient.


ireneisdoomed:
The solution to iterate the evidence file to look for duplicates has been removed.

The problem was that the rows were indeed exploded when the merge was done. I have simply dropped the duplicates for those rows.

modules/PheWAS.py
```diff
@@ -127,58 +157,76 @@ def main():
...
+            catalog.dropna(subset=["gene"], inplace=True)
+            # Merging dataframes: more records due to 1:many associations
+            enriched_catalog = pd.merge(catalog, phewas_w_consequences, on=["gene", "snp"])
+
+            one2many_variants = {i for i in phewas_w_consequences["snp"] if phewas_w_consequences["snp"].tolist().count(i) > 1}
+            enriched_catalog["variant_id"] = enriched_catalog.dropna(how="any", subset=["chrom", "pos", "ref", "alt"]).apply(lambda X: write_variant_id(X, one2many_variants), axis=1)
+            enriched_catalog.drop(["csq_arr", "most_severe_gene_csq"], axis=1, inplace=True)
+
+            # Dropping exploded duplicates
+            columns = ['chromosome', 'basepair', 'gene','snp', 'phewas_code', 'phewas_string', 'cases', 'odds_ratio','p','chrom','pos','most_severe_csq','consequence_link']
```
AsierGonzalez, line 174: This does not consider rsids with different positions or functional consequences as duplicates. However, given than the `one2many_variants` dictionary is only built for those ids, if that ever happened, their `variant_id` would be `None` so they would all share the same `variant_id` in the UAFs. If that happened, there would still be duplicates.

This situation is theoretically possible but I don't know if it happens with this data, so the question is whether it's worth putting the necessary safeguards in place to avoid the issues.

AsierGonzalez (commented):
The changes look good in general. I just have some comments that it would be good if you could answer but they would not block the merge.
The most notable thing I have found is that during duplicate removal, the code does not seem to consider the case in which the same rsid may have different positions or functional consequences (see comment in row 174). If that happened, we would still have duplicates. I admit that although this is theoretically possible, it may not be an issue with PheWAS data so I don't know if we want to control something that may never happen.

modules/PheWAS.py
AsierGonzalez: After discussions, we decided no t to code for this situation as it isn't expected to happen in the PheWAS data and if it does the duplicates will arise when running the validator or the pipeline

AsierGonzalez (dismissed):

AsierGonzalez (approved):

PR #51, opened 2020-11-16
[il-GEL] Updated script

ireneisdoomed:
As discussed on Friday 13th, four changes have been made to the current Panel App parser:

1. `evidence.allelic_requirement` has been added;
2. `panel_id, panel_name, panel_version` have been changed to `study_id, study_name, study_version`. In addition, these fields are not only in `unique_association_fields` but also can be found under `evidence`;
3. `disease.source_name` now shows the phenotype of the evidence, not the original string;
4. `evidence.phenotypes_array` has been added, as an array listing all the phenotypes present in the source string

[This is an example](https://github.com/opentargets/evidence_datasource_parsers/files/5546478/test-results.json.gz) of a newly generated string.

Thinking about possible incompatibilities with FE, please see if we should limit the `evidence.allelic_requirement` string.

A small set of strings have been generated and validated. The new pipeline is currently running.

### DIFF
--- a/modules/GenomicsEnglandPanelApp.py
+++ b/modules/GenomicsEnglandPanelApp.py
@@ -320,6 +320,8 @@ class PanelApp_evidence_generator():
                     'resource_score' : resource_score,
                     'urls' : urls,
                     'confidence' : self.evidence_classification,
+                    'panel_id': self.panel_id,
+                    'panel_version': self.panel_version,
                     'allelic_requirement' : self.mode_of_inheritance
                     }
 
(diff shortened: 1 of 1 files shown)


d0choa:
the content of the evidence looks great. It has everything we need. If somebody else can look at the code, we are ready to go.

AsierGonzalez:
What do you mean here, @ireneisdoomed?

> Thinking about possible incompatibilities with FE, please see if we should limit the evidence.allelic_requirement string.

ireneisdoomed:
I was just saying that there are strings, like the example I have attached, where this string is quite long and comprehensive. Was hesitant if these would look nice in the platform.

d0choa:
@ireneisdoomed can you add the most extreme cases here, and a sense of how common they are? It will help quite a loton specifying the FE task #1234

ireneisdoomed:
Sure. Will do.

DSuveges (approved):

PR #52, opened 2020-11-18
Ag crispr cell lines

AsierGonzalez:
Updates to CRISRP module to add and array with cell lines to `evidence.cell_lines`. The module had to be refactored to use `python_jsonsschema_objects` instead of the Open Targets `data_model` library.


DSuveges (approved):
All looks good.

PR #56, opened 2021-01-04
Solr phenodigm 1904

AsierGonzalez:
Merge working version of the MouseModels parser


AsierGonzalez:
Apart from the code to run the `MouseModels.py` parser, the PR also includes an updated README with all the information about running the parser and typo fixes in the information about the CRISPR and OT genetics parsers.

README.md
```diff
@@ -148,19 +148,70 @@ The parser has five parameters but only two of them are compulsory, the other th
...
+# Set environment variables
+export PYTHONPATH=.
+```
+Additionally, the parser needs access to Google Cloud, which requires downloading a key JSON file and setting the `GOOGLE_APPLICATION_CREDENTIALS` variable. Ask someone from the back-end team about it.
+
+#### Running MouseModels in a Google Cloud virtual machine
+
+Start the machine and connect to it via SSH.
+
+Before you start you may want to change the name of the output JSON by editing `MOUSEMODELS_EVIDENCE_FILENAME` variable in [settings.py](https://github.com/opentargets/evidence_datasource_parsers/blob/solr_phenodigm_1904/settings.py):
```
ireneisdoomed, line 188:
```suggestion
Before you start you may want to change the name of the output JSON by editing `MOUSEMODELS_EVIDENCE_FILENAME` variable in [settings.py](https://github.com/opentargets/evidence_datasource_parsers/blob/solr_phenodigm_1904/settings.py):
```

I would suggest erasing this comment and change the Global variable in `settings.py` to include the current date automatically in this way, for example:

`MOUSEMODELS_EVIDENCE_FILENAME = f"phenodigm-{datetime.today().strftime('%Y-%m-%d')}.json"`

This would output, as of today: `phenodigm-2021-01-05.json`

settings.py
```diff
@@ -101,10 +104,12 @@ class Config:
...
-    # mouse models
-    MOUSEMODELS_PHENODIGM_SOLR = 'http://localhost:8983' # 'solrclouddev.sanger.ac.uk'
-    # TODO remove refs to user directories
-    MOUSEMODELS_CACHE_DIRECTORY = '.phenodigmcache'
+    # mouse models (Phenodigm)
+    # used to be 'http://localhost:8983' # 'solrclouddev.sanger.ac.uk'
+    MOUSEMODELS_PHENODIGM_SOLR = 'http://www.ebi.ac.uk/mi/impc'
+    # write to the cloud direcly
+    MOUSEMODELS_CACHE_DIRECTORY = 'PhenoDigm/phenodigmcache'
+    MOUSEMODELS_EVIDENCE_FILENAME = 'phenodigm-4-4-2019.json'
```
ireneisdoomed, line 112: As suggested in the previous comment, I would change this variable to:
`MOUSEMODELS_EVIDENCE_FILENAME = f"phenodigm-{datetime.today().strftime('%Y-%m-%d')}.json"`

AsierGonzalez: I would prefer to remove it from the settings file altogether and convert it into a CLI parameter as I have done in the ClinGen, Gene2Phenotype and CRISPR parsers but I accept that at least this solution is better than having it hardcoded. The only problem happens if the parser is run twice on the same day and both outputs should be kept, as the second run would replace the output of the first one. However, I don't expect this to happen very often. Something to keep in mind when for future changes.

ireneisdoomed (approved):

PR #59, opened 2021-01-19
Ag clingen new schema 1278

AsierGonzalez:
Revamp of the ClinGen parser to align evidence strings with new schema. This is the summary of the changes made:

- JSON schema is **NOT** loaded
- Gene symbols are **NOT** mapped to Ensembl
- Evidence strings are stored as dictionaries and serialised as JSON
- The name of Gene Curation Expert Panel (`GCEP` column) is stored in the `studyId` field aking to what is done in Gene2Phenotype or Genomics England PanelApp evidence strings.
- The report url needed to link to ClinGen's website is stored in `urls`.
- The scoring based on the classification/confidence has been removed.

### DIFF
--- a/modules/ClinGen.py
+++ b/modules/ClinGen.py
@@ -8,15 +8,6 @@ import pandas as pd
 import requests
 import argparse
 
-ClinGen_classification2score = {
-    "Definitive": 1,
-    "Strong": 1,
-    "Moderate": 0.5,
-    "Limited": 0.01,
-    "Disputed": 0.01,
-    "Refuted": 0.01,
-    "No Reported Evidence": 0.01,
-}
 
 class ClinGen():
     def __init__(self):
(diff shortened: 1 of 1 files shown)


DSuveges:
I pulled the branch, runs just fine. Number of evidence: 1189, number of unmapped disease terms: 8.

The generated json strings successfully validated against the json schema:

```bash
opentargets_validator \
    --schema https://raw.githubusercontent.com/opentargets/json_schema/master/opentargets.json  \
    clingen_evid.json
```

modules/ClinGen.py
```diff
@@ -89,165 +47,78 @@ def process_gene_validity_curations(self, in_filename, out_filename, unmapped_di
...
-        with open(filename) as f:
-            f.readline() # Don't do anything with first row
-            second_row = f.readline()
-            file_created_date = second_row.split(": ")[1].strip() # Extract date and clean it
-
-
+        # When reading csv file skip header lines that don't contain column names
         gene_validity_curation_df = pd.read_csv(filename, skiprows= [0,1,2,3,5], quotechar='"')
         for index, row in gene_validity_curation_df.iterrows():
             print("{} - {}".format(row["GENE SYMBOL"], row["DISEASE LABEL"]))
```
DSuveges, line 53: Do we want to keep print statements? Wouldn't make more sense to have log statements instead? We can also do an other round of iteration when when we synchronise all logging, and print.

DSuveges (approved):
Works fine, output as expected. Updates in the code make sense.

modules/ClinGen.py
AsierGonzalez: You're right, that's a mistake - I'll fix it.

AsierGonzalez: You're right, that's a mistake - I'll fix it.

PR #60, opened 2021-01-19
[Il progeny rewrite] #1283

ireneisdoomed:
Revamp of the PROGENy parser to align evidence strings with new schema. This is the summary of the changes made:

- We no longer use the Open Targets library to build the schema.
- JSON schema is NOT loaded
- The disease and pathway mappings have been removed from the code. This is now imported from `pathway2Reactome_mappings.tsv` and `cancer2EFO_mappings.tsv `in the resources folder. 
- The input file has been removed from resources.
- PROGENy has been added as a data source to the README.
- References to the PROGENy files in `settings.py` have been removed.

As pointed out by @AsierGonzalez, there are still errors affecting the pathway mappings. This is partially described in [#713](https://github.com/opentargets/platform/issues/713) and has yet to be addressed


resources/pathway2Reactome_mappings.tsv
```diff
@@ -0,0 +1,15 @@
+pathway	target	reactomeId	description
+Androgen	AR	R-HSA-8940973	RUNX2 regulates osteoblast differentiation
+EGFR	EGFR	R-HSA-8856828	Clathrin-mediated endocytosis
+Estrogen	ESR1	R-HSA-8939902	Regulation of RUNX2 expression and activity
+Hypoxia	HIF1A	R-HSA-123456	XXX Pathway desc to be updated
```
AsierGonzalez, line 5: Can the description be updated? `R-HSA-123456` seems to be a placeholder and I couldn't easily find an alternative match. If not, should it be removed?

resources/pathway2Reactome_mappings.tsv
```diff
@@ -0,0 +1,15 @@
...
+EGFR	EGFR	R-HSA-8856828	Clathrin-mediated endocytosis
+Estrogen	ESR1	R-HSA-8939902	Regulation of RUNX2 expression and activity
+Hypoxia	HIF1A	R-HSA-123456	XXX Pathway desc to be updated
+JAK.STAT	JAK1, JAK2, STAT1, STAT2	R-HSA-6785807	Interleukin-4 and 13 signaling
+MAPK	PRKMK1, MAP2K2, RAF1	R-HSA-2559580	Oxidative Stress Induced Senescence
+NFkB	NFKB1, TLR4, RELA	R-HSA-9020702	Interleukin-1 signaling
+PI3K	PIK3CA, PIK3CA	R-HSA-8853659	RET signaling
+TGFb	TGFBR1, TGFBR2	R-HSA-2173788	Down regulation of TGF-beta receptor signaling
+TNFa	TNFRSF1A	R-HSA-5357956	TNFR1-induced NFkappaB signaling pathway
+Trail	TNFSF10, BCL2, BCL2L1, BCL2L2, MCL1	R-HSA-123456	XXX Pathway desc to be updated
```
AsierGonzalez, line 12: Same as above, can the description be updated? R-HSA-123456 seems to be a placeholder and I couldn't easily find an alternative match. If not, should it be removed?

resources/cancer2EFO_mappings.tsv
```diff
@@ -0,0 +1,75 @@
+Cancer_type_acronym	Cancer_type_name	EFO_id	EFO_name	Source
```
AsierGonzalez, line 1: Is the `Source` column needed? If so, does it need to be an URL?

modules/PROGENY.py
```diff
@@ -1,234 +1,140 @@
...
+        # Create spark session     
+        self.spark = SparkSession.builder \
+                .appName('progeny') \
+                .getOrCreate()
+
+        # Initialize mapping variables
+        self.mappingStep = mappingStep
+    
+        # Initialize input files
+        self.inputFile = inputFile
```
AsierGonzalez, line 21: Same as above: does this need to be an object attribute? It could be passed to directly to `writeEvidenceFromSource()` which is where it's used

modules/PROGENY.py
```diff
@@ -1,234 +1,140 @@
...
+
+class progenyEvidenceGenerator():
+    def __init__(self, inputFile, mappingStep):
+        # Create spark session     
+        self.spark = SparkSession.builder \
+                .appName('progeny') \
+                .getOrCreate()
+
+        # Initialize mapping variables
+        self.mappingStep = mappingStep
```
AsierGonzalez, line 18: Does this need to be an object attribute? It could be passed to directly to `writeEvidenceFromSource()` which is where it's used

modules/PROGENY.py
```diff
@@ -1,234 +1,140 @@
...
+        '''
+        # Read input file
+        self.dataframe = self.spark.read \
+                                .option("header", "true") \
+                                .option("delimiter", "\t") \
+                                .option("inferSchema", "true") \
+                                .csv(self.inputFile)
+
+        # Mapping step
+        if self.mappingStep:
```
AsierGonzalez, line 38: Why would you want to skip the disease mapping? To leave it to the ETL to do it?

modules/PROGENY.py
```diff
@@ -1,234 +1,140 @@
...
+        return self.dataframe
+
+    @staticmethod
+    def parseEvidenceString(row):
+        try:
+            evidence = {
+                "datasourceId" : "progeny",
+                "datatypeId" : "affected_pathway",
+                "diseaseFromSource" : row["Cancer_type"],
+                "diseaseFromSourceMappedId" : row["EFO_id"],
```
AsierGonzalez, line 93: What is the value of `row["Cancer_type"]` and `row["EFO_id"]` when `mappingStep` is `False`?

modules/PROGENY.py
```diff
@@ -1,234 +1,140 @@
...
+                .getOrCreate()
+
+        # Initialize mapping variables
+        self.mappingStep = mappingStep
+    
+        # Initialize input files
+        self.inputFile = inputFile
+        self.dataframe = None
+
+    def writeEvidenceFromSource(self):
```
AsierGonzalez, line 24: It does not write the evidence anywhere, it returns a variable. Could it be clearer to call it something like `processEvidenceFromSource`? Similar names are used elsewhere (see the [ClinGen](https://github.com/opentargets/evidence_datasource_parsers/blob/master/modules/ClinGen.py#L37) or [Gene2Phenotype](https://github.com/opentargets/evidence_datasource_parsers/blob/master/modules/Gene2Phenotype.py#L239) parsers for examples)

AsierGonzalez (commented):
I have some questions, please see inline comments.

resources/cancer2EFO_mappings.tsv
ireneisdoomed: I think it is good to include the source for an easier understanding in the future. But yes, I can definitely change the URLs to a description.

resources/pathway2Reactome_mappings.tsv
ireneisdoomed: This sort of errors were previously described in #713 and have yet to be addressed. I'm making a comment in the PR for clarification.

resources/pathway2Reactome_mappings.tsv
ireneisdoomed: This sort of errors were previously described in #713 and have yet to be addressed. I'm making a comment in the PR for clarification.

modules/PROGENY.py
ireneisdoomed: Yes.

modules/PROGENY.py
ireneisdoomed: Changed it to `generateEvidenceFromSource`

modules/PROGENY.py
ireneisdoomed: The `diseaseFromSource` field will remain the same, if the mappingStep is False only `diseaseFromSourceMappedId` will be affected. I've added the condition that if that were the case, this will be `None` as a safeguard. Until we don't switch off the mappingStep on our side, we don't know what is the best solution to feed the ETL.

AsierGonzalez (dismissed):
Thanks @ireneisdoomed!

modules/PROGENY.py
ireneisdoomed: This approach was improved by creating the field only if the value of `row["EFO_id"]` existed.

README.md
```diff
@@ -184,6 +184,26 @@ This is how it is run only specifying the required parameters:
...
+- cancer2EFO_mappings.tsv: File containing the mappings between the acronym of the type of cancer and its respective disease listed in EFO. This file can be found in the `resources` directory.
+- pathway2Reactome_mappings.tsv: File containing the mappings between the analysed pathway and their respective target and ID in Reactome. This file can be found in the `resources` directory.
+
+The source table is then formatted into a compressed set of JSON lines following the schema of the version to be used.
+
+The parser requires three parameters:
+- `-i`, `--inputFile`: Name of tsv file located in the [PROGENy bucket](https://storage.googleapis.com/otar000-evidence_input/PROGENy/data_files/progeny_normalVStumor_opentargets.txt).
+- `-o`, `--outputFile`: Name of the evidence compressed JSON file containing the evidence strings.
+- `-s`, `--mappingStep`: State whether to run the disease to EFO term mapping step. It is True by default.
```
AsierGonzalez, line 200: This is now called `skipMapping` instead of `mappingStep`

modules/PROGENY.py
```diff
@@ -1,234 +1,144 @@
...
-
-
+    # Initiating parser
+    parser = argparse.ArgumentParser(description=
+    "This script generates evidences for the PROGENy data source.")
+
+    parser.add_argument("-i", "--inputFile", required=True, type=str, help="Input source .txt file.")
+    parser.add_argument("-o", "--outputFile", required=True, type=str, help="Name of the evidence compressed JSON file containing the evidence strings.")
+    parser.add_argument("-s", "--skipMapping", required=False, action="store_true", help="State whether to skip the disease to EFO mapping step.")
+    parser.add_argument("-l", "--logFile", help="Destination of the logs generated by this script.", type=str, required=False)
```
AsierGonzalez, line 112: Add info about this option to README

modules/PROGENY.py
```diff
@@ -1,234 +1,144 @@
...
+        diseaseMappingsFile = (self.spark
+                        .read.csv("resources/cancer2EFO_mappings.tsv", sep=r'\t', header=True)
+                        .select("Cancer_type_acronym", "EFO_id")
+                        .withColumnRenamed("Cancer_type_acronym", "Cancer_type"))
+
+        self.dataframe = self.dataframe.join(
+            diseaseMappingsFile,
+            on="Cancer_type",
+            how="inner"
+)
```
AsierGonzalez, line 60: Parenthesis is a not aligned with code here

modules/PROGENY.py
```diff
@@ -1,234 +1,144 @@
...
+            evidence = {
+                "datasourceId" : "progeny",
+                "datatypeId" : "affected_pathway",
+                "resourceScore" : row["P.Value"],
+                "pathwayName" : row["description"],
+                "pathwayId" : row["reactomeId"],
+                "targetFromSourceId" : row["target"],
+                "diseaseFromSource" : row["Cancer_type"]
+            }
+            if "EFO_id" in row:
```
AsierGonzalez, line 98: Nice, avoids issue when mapping is skipped. If you want to be extra secure you can also check that the value of the field is not `None` but it isn't really needed.

AsierGonzalez (changes requested):
Just some minimal changes needed.

ireneisdoomed:
New changes:
- more logging messages
- all input files are set as an argument
- mapping step has been revamped to skipMapping
- I have added some logic outside of the parsing function to remove the fields which are empty when the disease mapping step is skipped
- README has been updated in consonance

DSuveges (approved):

PR #62, opened 2021-01-19
[Il slapenrich rewrite] #1282

ireneisdoomed:
Revamp of the SLAPenrich parser to align evidence strings with new schema. This is the summary of the changes made:

- JSON schema is NOT loaded
- The disease mappings have been removed from the code. This is now imported from `cancer2EFO_mappings.tsv` in the resources folder.
- The input file has been removed from resources.
- SLAPenrich has been added as a data source to the README.
- References to the SLAPenrich files in settings.py have been removed.


modules/SLAPEnrich.py
```diff
@@ -1,287 +1,117 @@
...
+    # Parsing parameters
+    args = parser.parse_args()
+
+    inputFile = args.inputFile
+    outputFile = args.outputFile
+    mappingStep = args.mappingStep
+
+    # Initialize logging:
+    logging.basicConfig(
+    filename='evidence_builder.log',
```
DSuveges, line 98: The parsers are expected to log to the standard error, right?

modules/SLAPEnrich.py
```diff
@@ -1,287 +1,117 @@
...
+        return self.dataframe
+
+    @staticmethod
+    def parseEvidenceString(row):
+        try:
+            evidence = {
+                "datasourceId" : "slapenrich",
+                "datatypeId" : "affected_pathway",
+                "diseaseFromSource" : row["Cancer_type_acronym"],
+                "diseaseFromSourceMappedId" if row["EFO_id"] else None : row["EFO_id"],
```
DSuveges, line 69: I'm not sure if this works as expected. I'll look into this.

modules/SLAPEnrich.py
```diff
@@ -1,287 +1,117 @@
...
+                "resourceScore" : row["pval"],
+                "pathwayName" : row["pathwayDescription"],
+                "pathwayId" : row["pathwayId"],
+                "targetFromSourceId" : row["gene"]
+            }
+            evidence.pop(None, None)
```
DSuveges, line 75: What does this pop do?

DSuveges: I see: if `row["EFO_id"] is None` an extra key/value pair (`None:None`) is added to the dictionary. In which case you need to pop that key. But at least the `"diseaseFromSourceMappedId"` will not be there with the `None` value. 

I would consider doing:
```Python
evidence = {
    "datasourceId" : "slapenrich",
    "datatypeId" : "affected_pathway",
    "diseaseFromSource" : row["Cancer_type_acronym"],
    "resourceScore" : row["pval"],
    "pathwayName" : row["pathwayDescription"],
    "pathwayId" : row["pathwayId"],
    "targetFromSourceId" : row["gene"]
}
if row['EFO_id]:
   evidence['diseaseFromSourceMappedId'] = row['EFO_id]
```

But I think if you use pyspark's json export methods, keys with `None` values will not be written into the JSON.

README.md
```diff
@@ -184,6 +184,25 @@ This is how it is run only specifying the required parameters:
...
+
+- progeny_normalVStumor_opentargets.txt: Main file that contains a systematic comparison of on somatic mutations from TCGA across 25 different cancer types and a collection of pathway gene sets from Reactome. This file can be downloaded [here](https://storage.googleapis.com/otar000-evidence_input/SLAPEnrich/data_file/slapenrich_opentargets-21-12-2017.tsv) from the _otar000-evidence_input_ bucket.
+- cancer2EFO_mappings.tsv: File containing the mappings between the acronym of the type of cancer and its respective disease listed in EFO. This file can be found in the `resources` directory.
+
+The source table is then formatted into a compressed set of JSON lines following the schema of the version to be used.
+
+The parser requires three parameters:
+- `-i`, `--inputFile`: Name of tsv file located in the [SLAPEnrich bucket](https://storage.googleapis.com/otar000-evidence_input/SLAPEnrich/data_file/slapenrich_opentargets-21-12-2017.tsv).
+- `-o`, `--outputFile`: Name of the evidence compressed JSON file containing the evidence strings.
+- `-s`, `--mappingStep`: State whether to run the disease to EFO term mapping step. It is True by default.
```
DSuveges, line 198: I think the mapping step is `-m`.

modules/SLAPEnrich.py
```diff
@@ -1,287 +1,117 @@
...
-if __name__ == "__main__":
-    main()
-
+    # Initiating parser
+    parser = argparse.ArgumentParser(description=
+    "This script generates evidences for the SLAPEnrich data source.")
+
+    parser.add_argument("-i", "--inputFile", required=True, type=str, help="Input source .tsv file.")
+    parser.add_argument("-o", "--outputFile", required=True, type=str, help="Name of the evidence compressed JSON file containing the evidence strings.")
+    parser.add_argument("-m", "--mappingStep", required=False, type=bool, default=True, help="State whether to run the disease to EFO term mapping step.")
```
DSuveges, line 87: I think it doesn't work. In this format, `-m` expects an argument, however, you only use it as a flag. 

Write something like this:

```python
parser.add_argument('-m', action='store_false') # if the flag is present, it's value will be false
```

DSuveges: Also if it made the modification to be able to run the script without mapping, the script fails.

modules/SLAPEnrich.py
AsierGonzalez: I like it but it would still fail when using `--mappingStep` because `row['EFO_id']` would not be defined, right? So peharps a `if 'EFO_id' in row:` is needed?

DSuveges: I have to think about... the json generation should be the very last step, this decision should happen before. But I might misunderstood.

ireneisdoomed:
New changes:
- more logging messages
- all input files are set as an argument
- mapping step has been revamped to `skipMapping`
- I have added some logic outside of the parsing function to remove the fields which are empty when the disease mapping step is skipped
- README has been updated in consonance

DSuveges (approved):
All looks good.

PR #64, opened 2021-01-22
Ag g2p new schema 1279

AsierGonzalez:
Revamp of the Gene2Phenotype parser to align evidence strings with new schema. This is the summary of the changes made:

- JSON schema is NOT loaded
- Evidence strings are stored as dictionaries and serialised as JSON
- The scoring based on the classification/confidence has been removed.
- The mutation consequence is stored in `variantFunctionalConsequenceId`

Closes opentargets/platform#1279


AsierGonzalez:
Closes [#1279](https://github.com/opentargets/platform/issues/1279)

DSuveges (approved):
All looks good. I only made a bit update in the README and removed the `config.py` bits. That part will be phased out. At some point we'll need to update the way the logging is implemented.

AsierGonzalez:
Thank you @DSuveges!

PR #69, opened 2021-02-02
Il phewas rewrite

ireneisdoomed:
Revamp of the PROGENy parser to align evidence strings with new schema. This is the summary of the changes made:

- Implemented argparse with every input file as an argument.
- Better logging system.
- For those SNPs that map to different variants, instead of building a list with them and checking if the snp is in the list when I am building `variantId`, I built the list and dropped the rows in `phewasWithConsequences` containing those snps. Then when merging the data, a left join is performed, not an inner one.
- More detailed README.
- References to the PheWAS files in settings.py have been removed.


ireneisdoomed:
Example of the evidence:

```
{
    "datasourceId": "phewas_catalog",
    "datatypeId": "genetic_association",
    "diseaseFromSource": "Hemorrhoids",
    "diseaseFromSourceId": "455",
    "diseaseFromSourceMappedId": "EFO_0009552",
    "oddsRatio": "0.9181",
    "resourceScore": "0.03348",
    "studyCases": "2796",
    "targetFromSourceId": "ENSG00000149571",
    "variantFunctionalConsequenceId": "SO_0001627",
    "variantId": "11_126426694_C_T",
    "variantRsId": "rs4935969"
}
```

DSuveges (approved):
All looks good. Generated evidence strings pass validation.

PR #70, opened 2021-02-12
[pathways fix] 

ireneisdoomed:
Changes to the SLAPEnrich and PROGENy parsers after the implementation of the pathways field into the schema described in [#1387](https://github.com/opentargets/platform/issues/1387).


DSuveges (approved):
All looks good.

PR #74, opened 2021-03-03
Snakemake scaffold

ireneisdoomed:
Branch containing all the details related to the Snakemake implementation. 
- There are minor changes in some parsers to align with the snake file.
- I've deleted from resources the crispr input files and some legacy panel ID mappings from Panel App
- Should we move `resources/cancer2EFO_mappings.tsv` and `resources/pathway2Reactome_mappings.tsv` to the [mappings](https://github.com/opentargets/mappings) repo?
- Shall we remove the requirements.txt file? Environment is now read from Conda-env.yml
- What do you think about `utils/get_phenodigm_score_distribution.py`?


Snakefile
```diff
@@ -0,0 +1,465 @@
...
+        GS.remote(logFile)
+    shell:
+        """
+        cp  {input.diseaseMapping} {output.diseaseMapping}
+        cp  {input.pathwayMapping} {output.pathwayMapping}
+        gsutil cp {input.inputFile} {output.inputFile}
+        """
+
+## fetchSysbio           : fetches evidence data and study level information from GS
+rule fetchSysbio:
```
tskir, line 414: I notice there are pairs of steps, such as `sysbio` / `fetchSysbio`. Can't we unify them for simplicity? Or is there some reason I don't yet understand?

common/HGNCParser.py
```diff
@@ -83,7 +83,7 @@ def __init__(self):
 
     def _get_hgnc_data_from_json(self):
 
-        r = requests.get(Config.GENES_HGNC)
+        r = requests.get('http://ftp.ebi.ac.uk/pub/databases/genenames/new/json/hgnc_complete_set.json')
```
tskir, line 86: Could we also import this from a config file instead of hardcoding?

tskir: Ah, I see the config file is a YAML. Then perhaps not importing, but passing this as a command line parameter? Anyway, I don't think hardcoding is the best way to go

envs/conda-env.yml
```diff
@@ -0,0 +1,132 @@
+name: snakemake
+channels:
+  - anaconda
+  - conda-forge
+  - defaults
+dependencies:
```
tskir, line 6: That's an impressive list of dependencies, both Conda and PIP. I assume that it was generated by `conda list --export` + `pip freeze` (or something along those lines). Pinning all of them to specific versions has advantages; for one, it ensures maximum reproducibility. However, do we have a plan on how to re-generate/update this list in the future?

envs/conda-env.yml
```diff
@@ -0,0 +1,132 @@
...
+    - ratelimiter==1.2.0.post0
+    - smart-open==4.2.0
+    - smmap==3.0.5
+    - snakemake==6.0.0
+    - toposort==1.6
+    - traitlets==5.0.5
+    - typing-extensions==3.7.4.3
+    - wrapt==1.12.1
+    - zipp==3.4.0
+prefix: /Users/irene/miniconda3/envs/snakemake
```
tskir, line 131: I'm assuming the path from your home directory shouldn't be in the repository :)

tskir (commented):
Overall I like this very much! I've never worked with Snakemake before, and its syntax appears to be even more concise than that of Nextflow. Very nice.

ireneisdoomed:
The current version is fully functional, at least locally, except for two parsers: EPMC and Phenodigm, the latter waiting to be merged with Kirill's work. As for EPMC, I have encountered a problem when downloading the parquet files directory, it seems that snakemake may have problems with the GCP folder system.

We have duplicity in the function of several files: `environment.yaml`, `envs/conda-env.yml`, `requirements.txt`. In my opinion, as soon as EPMC and Phenodigm are integrated into the pipeline, we could merge the branch and continue iterating from there - or do you think it is better to keep this branch open until we are fully satisfied?

 As future improvements:
- Make it compatible with Python 3.9.
- Unify the declaration of dependencies in a single file.
- Give a twist to storing all files in `tmp`, as Snakemake creates a temporary local directory following the structure of the remote directory. We could save the step of copying all the files to `tmp`.
- Review the satellite files usefulness (`ontology_config.ini`, .`gitignore`,...).
- Update the README
- Dockerise the module
- ...

Snakefile
```diff
@@ -0,0 +1,465 @@
+from datetime import datetime
+from snakemake.remote.GS import RemoteProvider as GSRemoteProvider
+GS = GSRemoteProvider()
+
+# --- Settings --- #
+# Current time in YYYY-MM-DD-hh-mm format:
```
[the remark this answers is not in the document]
tskir: I see that the comment was updated, but I think that it now disagrees with the code:

* `# Current time in YYYY-MM-DD-hh-mm format:`
* `timeStamp = datetime.now().strftime("%Y-%m-%d")`

Snakefile
```diff
@@ -0,0 +1,465 @@
...
+timeStamp = datetime.now().strftime("%Y-%m-%d")
+
+# Configuration is read from the config yaml:
+configfile: 'configuration.yaml'
+
+# Schema version is now hardcoded, version will be read from the command line later:
+schemaFile = config['global']['schema_file']
+logFile = f"{config['global']['logDir']}/evidence_parser.{timeStamp}.log"
```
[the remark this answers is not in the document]
tskir: @ireneisdoomed Do you suppose we could add this to the list of follow-up actions for the next ticket?

Snakefile
```diff
@@ -0,0 +1,465 @@
...
+        """
+        curl  {params.diseaseMapping} > {output.diseaseMapping}
+        gsutil cp {input.inputFile} {output.inputFile}
+        gsutil cp {input.consequencesFile} {output.consequencesFile}
+        """
+
+## fetchSlapenrich           : fetches SLAPenrich table from GS
+rule fetchSlapenrich:
+    input:
+        inputFile=GS.remote(f"{config['SLAPEnrich']['inputBucket']}/slapenrich_opentargets-21-12-2017.tsv")
```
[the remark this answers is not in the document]
tskir: @ireneisdoomed What do you think about this one? Also could be moved to a future ticket if necessary

Snakefile
tskir: @ireneisdoomed What do you think? Should we add that to the follow-up ticket or is this not worth the effort at all?

common/HGNCParser.py
tskir: @ireneisdoomed What about this one?

tskir (commented):
@ireneisdoomed Thank you for the updates!

And regarding the development approach: I don't mind at all merging this PR once you consider it done. However, before that I think it's a good idea to resolve all open discussions (by either addressing them, closing as irrelevant, or explicitly adding to the list of future improvements)—otherwise we risk losing them

Snakefile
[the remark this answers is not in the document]
ireneisdoomed: Changed in `a3b491a`

Snakefile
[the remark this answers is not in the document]
ireneisdoomed: Changed in `a3b491a`. I have removed this variable as it is no longer an input of any of the parsers.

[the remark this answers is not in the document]
tskir: Cool! This thread was less about that particular variable and more about the name styles in general :) But as I said, this is not something really important anyway.

common/HGNCParser.py
DSuveges: +1 for adding that file as command line argument. The current implementation makes our data sources obscured.

common/HGNCParser.py
```diff
@@ -83,7 +83,7 @@ def __init__(self):
...
         data = r.json()
 
         for row in tqdm(data['response']['docs'],
```
DSuveges, line 89: Please remove tqdm from this parser. This should be trivial, without any implication.

envs/conda-env.yml
DSuveges: At some point we certainly need to do something about this list for sure. However this is not an urgent fix, I would rather open a ticket with a list of issues we need to resolve later. I see this is not a blocker for the merge.

modules/MouseModels.py
```diff
@@ -1,6 +1,5 @@
 import requests
 import collections
-from tqdm import tqdm
```
DSuveges, line 3: @tskir , what are the plans: are we rolling out the `PhenoDigm.py` parser into production for this release? If yes, we can remove this file. If no, we can keep it for one more release.

settings.py
```diff
@@ -4,23 +4,6 @@
 import pkg_resources as res
```
DSuveges, line 4: Is any of our parsers using data in the `settings.py` file still? I feel we should remove this file, it's quite confusing having a `settings.py` and a `configuration.yaml` too. This should go.

configuration.yaml
```diff
@@ -0,0 +1,65 @@
+global:
+  logDir: gs://ot-team/irene/evidence_test
```
DSuveges, line 2: We should remove the explicit destination of the log files. The logfile can either be a file in the working directory or specified via a command line option.

Snakefile
```diff
@@ -0,0 +1,496 @@
...
+        GS.remote(f"{config['SLAPEnrich']['outputBucket']}/slapenrich-{timeStamp}.json.gz"),
+        GS.remote(f"{config['Gene2Phenotype']['outputBucket']}/gene2phenotype-{timeStamp}.json.gz"),
+        GS.remote(f"{config['CRISPR']['outputBucket']}/crispr-{timeStamp}.json.gz"),
+        GS.remote(f"{config['SysBio']['outputBucket']}/sysbio-{timeStamp}.json.gz"),
+        GS.remote(f"{config['Phenodigm']['outputBucket']}/phenodigm-{timeStamp}.json.gz")
+
+# --- Auxiliary Rules --- #
+## help               : prints help comments for Snakefile
+rule help:
+    input: "Snakefile"
```
DSuveges, line 27: This makes the pipeline run bound to the script directory. I think we should be more flexible than that. However this is not a blocker for merging.

Snakefile
```diff
@@ -0,0 +1,496 @@
...
+
+# --- Data sources parsers --- #
+## clingen          : processes the Gene Validity Curations table from ClinGen
+rule clingen:
+    input:
+        f"tmp/ClinGen-Gene-Disease-Summary-{timeStamp}.csv"
+    output:
+        evidenceFile=GS.remote(f"{config['ClinGen']['outputBucket']}/ClinGen-{timeStamp}.json.gz"),
+        unmappedDiseases=f"tmp/unmappedDiseases/clingen_unmapped_diseases-{timeStamp}.lst"
+    conda:
```
DSuveges, line 44: Why do we need to specify the environment all the time? I thought we have an environment described by the yaml, and you run the entire pipeline on that env. I think it would be important if certain rules would be different requirements. But I might be wrong.

Snakefile
```diff
@@ -0,0 +1,496 @@
...
+            modules/GeneticsPortal.py -- \
+            --locus2gene gs://genetics-portal-data/l2g/200127 \
+            --toploci gs://genetics-portal-data/v2d/200207/toploci.parquet \
+            --study gs://genetics-portal-data/v2d/200207/studies.parquet \
+            --threshold 0.05 \
+            --variantIndex gs://genetics-portal-data/variant-annotation/190129/variant-annotation.parquet  \
+            --ecoCodes gs://genetics-portal-data/lut/vep_consequences.tsv \
+            --outputFile gs://genetics-portal-analysis/l2g-platform-export/data/genetics_portal_evidence.json.gz
+
+        gcloud dataproc clusters delete \
```
DSuveges, line 85: The delete is not necessary: the cluster is deleted after 10 minutes idle time `--max-idle=10m `, but can stay.

Snakefile
DSuveges: This is an interesting question, when we were brainstorming about the pipeline, we often argued back and forth. Now, I'm not sure we need to do it at all. I'm working with some of the genetics portal pipelines, and there are some interesting implementations: in [this snakefile](https://github.com/opentargets/genetics-v2d-data/blob/master/snakefiles/study_and_top_loci_tables.Snakefile) input files are not explicitly fetched, instead, they just referred and snakemake handles in the background:

```python
from snakemake.remote.FTP import RemoteProvider as FTPRemoteProvider
from snakemake.remote.GS import RemoteProvider as GSRemoteProvider
from snakemake.remote.HTTP import RemoteProvider as HTTPRemoteProvider
```

I think we should do the same. @ireneisdoomed , please take a look at these files, without diving into them too deep, just for getting some inspiration. 

But again: if the current implementation works, it's not a dealbreaker. We'll collect these points under a ticket and we'll schedule a meeting to discuss priorities and what we can drop.

DSuveges (changes requested):

Snakefile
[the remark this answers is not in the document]
ireneisdoomed: I wanted to use the same pattern across all sources so we know the extension of the file just by looking at the Snakefile. It is true that as far as I know we don't expect this input file will change - same goes for the other ones like PheWAS. Do you think it is better to hardcode it?

common/HGNCParser.py
ireneisdoomed: The URL is specified in the yaml file + new parameter in the PheWAS parser.
Commit`7855a3d`

common/HGNCParser.py
ireneisdoomed: Done `d5f2178`

Snakefile
ireneisdoomed: It was just for peace of mind 😬

Snakefile
ireneisdoomed: I expected this to be necessary to address the particularities of the old Phenodigm environment. that's not the case anymore, so I've removed it.

Snakefile
ireneisdoomed: Addressed in `9725f11`

configuration.yaml
ireneisdoomed: New location for the logs: `otar000-evidence_input/parser_logs`

settings.py
ireneisdoomed: I absolutely agree, but it is still being used in some parts of the code: e.g. the Gene2Phenotype parser. 
Added to the follow-up ticket.

Snakefile
ireneisdoomed: Added to the follow-up ticket.

modules/MouseModels.py
ireneisdoomed: MouseModels.py has been removed.

envs/conda-env.yml
ireneisdoomed: Added to the follow-up ticket.

Snakefile
ireneisdoomed: Added to the follow-up ticket.

Snakefile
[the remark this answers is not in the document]
ireneisdoomed: Added to the follow-up ticket.

DSuveges (dismissed):

Snakefile
```diff
@@ -0,0 +1,448 @@
...
+    shell:
+        """
+        curl  {params.webSource_dd_panel} > {output.ddLocal}
+        gsutil cp {output.ddLocal} {output.ddBucket}
+        curl  {params.webSource_eye_panel} > {output.eyeLocal}
+        gsutil cp {output.eyeLocal} {output.eyeBucket}
+        curl  {params.webSource_skin_panel} > {output.skinLocal}
+        gsutil cp {output.skinLocal} {output.skinBucket}
+        curl  {params.webSource_cancer_panel} > {output.cancerLocal}
+        gsutil cp {output.cancerLocal} {output.cancerBucket}
```
tskir, line 348:
```suggestion
        curl {params.webSource_dd_panel} > {output.ddLocal}
        gsutil cp {output.ddLocal} {output.ddBucket}
        curl {params.webSource_eye_panel} > {output.eyeLocal}
        gsutil cp {output.eyeLocal} {output.eyeBucket}
        curl {params.webSource_skin_panel} > {output.skinLocal}
        gsutil cp {output.skinLocal} {output.skinBucket}
        curl {params.webSource_cancer_panel} > {output.cancerLocal}
        gsutil cp {output.cancerLocal} {output.cancerBucket}
```

Snakefile
```diff
@@ -0,0 +1,448 @@
...
+        diseaseMapping=config['PheWAS']['diseaseMapping'],
+    output:
+        inputFile=f"tmp/phewas_catalog-{timeStamp}.csv",
+        consequencesFile=f"tmp/phewas_w_consequences-{timeStamp}.csv",
+        diseaseMapping=f"tmp/phewascat_mappings-{timeStamp}.tsv"
+    log:
+        GS.remote(logFile)
+    shell:
+        """
+        curl  {params.diseaseMapping} > {output.diseaseMapping}
```
tskir, line 298:
```suggestion
        curl {params.diseaseMapping} > {output.diseaseMapping}
```

tskir (approved):
:tada:

PR #78, opened 2021-03-30
Genetics portal bugfix

DSuveges:
* Bugfix: `variantId` was not good: `alt_ref` changed to `ref_alt` 
* In the final json files there were some duplicates.  I removed them.


DSuveges:
Removing redundancy - interim solution which does not consider the p-value of the association:

```python
.dropDuplicates(['variantId', 'studyId', 'targetFromSourceId', 'diseaseFromSourceMappedId'])
```

Number of evidence strings: `393256` -> `393244` 
Number of unique fields: `393244` -> `393244`

So it seems things work well: the number of evidence strings slightly went down, but the number of unique evidence is the same. 

Newly updated evidence:
`gs://otar000-evidence_input/Genetics_portal/json/genetics_portal_evidence_2021.04.13`

ireneisdoomed (approved):
All fixed, thanks!

PR #79, opened 2021-04-08
Update Orphanet_573163 to EFO_1000453

ireneisdoomed:
The more specific mapping Orphanet_573163 is not included in EFO at the moment, hence leading to invalid evidence strings.
We are going back to the less specific mapping "Paraganglioma" until we know whether it is possible to bring Orphanet_573163 into EFO.


DSuveges (approved):
Looks good.

PR #81, opened 2021-04-21
EPMC parser

DSuveges:
Input: grounded cooccurrence file generated by Miguel.
Output: partitioned gzipped json

**Usage:**
```bash
python modules/EPMC.py \
    --cooccurrenceFile gs://ot-snapshots/literature/20210418/cooccurrences \
    --output  gs://ot-team/dsuveges/evidence_test/EPMC/epmc-2021-04-19 \
   --logFile epmc_parser.log \
   --local
```

Specifying `logFile` is optional, if not logs are written to standard error. `--local` is a flag effecting on how spark is initialised.


modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
+#!/usr/bin/env python
+
+import argparse
+import sys
+from pyspark import *
+from pyspark.sql import *
+from pyspark.sql.types import *
+from pyspark.sql.functions import *
+from pyspark.sql.window import Window
+import logging
```
tskir, line 10: As recommended by PEP8 (separating standard library and third party imports):

```suggestion
import argparse
import logging
import sys

from pyspark import *
from pyspark.sql import *
from pyspark.sql.types import *
from pyspark.sql.functions import *
from pyspark.sql.window import Window
```

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
+#!/usr/bin/env python
+
+import argparse
+import sys
+from pyspark import *
+from pyspark.sql import *
+from pyspark.sql.types import *
+from pyspark.sql.functions import *
```
tskir, line 8: Personally I find wildcard imports confusing and also just a tad dangerous; and Python documentation [also says they should be avoided](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package) (second to last paragraph)

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+from pyspark.sql.window import Window
+import logging
+
+
+# example run local, make sure you dont pass many parquet files as input to not run out of mem
+# also, double check you put the ram size properly in the local settings in the SparkConf map
+# python modules/EPMC.py --local \
+#        --cooccurrenceFile /home/mkarmona/src/opentargets/data/platform/epmc/epmc-cooccurrences \
+#        --outputFile test_out
+#        --logFile epmc_parser.log
```
tskir, line 18:
```suggestion
# Example for a local run. Make sure you don't pass many parquet files as input to avoid running out of memory.
# Also, double check you put the RAM size properly in the local settings in the SparkConf map.
# python modules/EPMC.py --local \
#        --cooccurrenceFile /home/mkarmona/src/opentargets/data/platform/epmc/epmc-cooccurrences \
#        --outputFile test_out
#        --logFile epmc_parser.log
```

tskir: Also, should this information perhaps be moved to the documentation (README)?

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+
+    ##
+    ## Parsing parameters:
+    ##
+    parser = argparse.ArgumentParser(description='This script generates target/disease evidence strings from ePMC cooccurrence files.')
+    parser.add_argument('--cooccurrenceFile', help='Partioned parquet file with the ePMC cooccurrences', type=str, required=True)
+    parser.add_argument('--outputFile', help='Resulting evidence file saved as compressed JSON.', type=str, required=True)
+    parser.add_argument('--logFile', help='Destination of the logs generated by this script.', type=str, required=False)
+    parser.add_argument('--local', help='Destination of the logs generated by this script.', action='store_true', required=False, default=False)
+    args = parser.parse_args()
```
tskir, line 31: Judgement call; I would move the parser definition from here into the `if __name__ == '__main__'` block. The reason is, when `parse_args()` is called in the main function, it makes it impossible to import from this module to anywhere. While we're not necessarily planning to import this `main()` function from anywhere, I think it's better for consistency to make it always importable

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+
+    # extract parameters:  
+    cooccurrenceFile = args.cooccurrenceFile
+
+
```
tskir, line 36:
```suggestion
    # Extract parameters.
    cooccurrenceFile = args.cooccurrenceFile

```

(extra blank line)

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+    # Parse output file:
+    out_file = args.outputFile
+
+    ##
+    ## Initialize spark session
+    ##
+    global spark
+    local = True
+
```
tskir, line 57: Global variables should generally be avoided; and in this case it's also not needed since the variable `spark` only exists within the `main()` function.

Also, the variable `local` appears to be not used.

```suggestion
```

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+        sparkConf = (SparkConf()
+                 .set("spark.driver.maxResultSize", "0")
+                 .set("spark.debug.maxToStringFields", "2000")
+                 .set("spark.sql.execution.arrow.maxRecordsPerBatch", "500000")
+                 )
+        spark = (
+            SparkSession.builder
+                .config(conf=sparkConf)
+                .getOrCreate()
+        )
```
tskir, line 82: Suggest using single quote string literals to align with the rest of the script:

```suggestion
    if args.local:
        sparkConf = (SparkConf()
                 .set('spark.driver.memory', '15g')
                 .set('spark.executor.memory', '15g')
                 .set('spark.driver.maxResultSize', '0')
                 .set('spark.debug.maxToStringFields', '2000')
                 .set('spark.sql.execution.arrow.maxRecordsPerBatch', '500000')
                 )
        spark = (
            SparkSession.builder
                .config(conf=sparkConf)
                .master('local[*]')
                .getOrCreate()
        )
    else:
        sparkConf = (SparkConf()
                 .set('spark.driver.maxResultSize', '0')
                 .set('spark.debug.maxToStringFields', '2000')
                 .set('spark.sql.execution.arrow.maxRecordsPerBatch', '500000')
                 )
        spark = (
            SparkSession.builder
                .config(conf=sparkConf)
                .getOrCreate()
        )
```

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+
+    ##
+    ## Log parameters:
+    ##
+    logging.info(f'Cooccurrence file: {cooccurrenceFile}')
+    logging.info(f'Output file: {out_file}')
+    logging.info(f'Generating evidence:')
+
+    filtered_target_terms = ["TEC", "TECS", "Tec", "tec", "'", "(", ")", "-", "-S", "S", "S-", "SS", "SSS",
+         "Ss", "Ss-", "s", "s-", "ss", "U3", "U6", "u6", "SNORA70", "U2", "U8"]
```
tskir, line 93: Perhaps this would be better defined as a module-level constant?

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+    filtered_cooccurrence_df = (
+        # Reading file:
+        spark.read.parquet(cooccurrenceFile)
+
+        # Filtering for diases/target cooccurrences:
+        .filter(
+            (col('type') == "GP-DS") & 
+            (col('isMapped') == True) & 
+            (col('pmid').isNotNull()) & 
+            (length(col('text')) < 600) &
```
tskir, line 109: The rest of the filters kind of make sense intuitively but not this one. What does it do? Why don't we want records with text longer than 600? And why exactly 600? :)

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+            (col('label1').isin(filtered_target_terms) == False)
+        )
+        # Renaming columns:
+        .withColumnRenamed("keywordId1", "targetFromSourceId")
+        .withColumnRenamed("keywordId2", "diseaseFromSourceMappedId")
+        .withColumnRenamed("label1", "targetFromSource")
+        .withColumnRenamed("label2", "diseaseFromSource")
```
tskir, line 116: To keep literals in the same style as the rest of the script:

```suggestion
        .withColumnRenamed('keywordId1', 'targetFromSourceId')
        .withColumnRenamed('keywordId2', 'diseaseFromSourceMappedId')
        .withColumnRenamed('label1', 'targetFromSource')
        .withColumnRenamed('label2', 'diseaseFromSource')
```

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+
+        .withColumn('tmp',col('pmid').cast(StringType()))
```
tskir, line 118:
```suggestion
        .withColumn('tmp', col('pmid').cast(StringType()))
```

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+    )
+
+    # Report on the number of diseases, targets and associations:
+    logging.info(f'Number of publications: {filtered_cooccurrence_df.select(col("tmp")).distinct().count()}')
+    logging.info(f'Number of targets: {filtered_cooccurrence_df.select(col("targetFromSourceId")).distinct().count()}')
+    logging.info(f'Number of diseases: {filtered_cooccurrence_df.select(col("diseaseFromSourceMappedId")).distinct().count()}')
+    logging.info(f'Number of associations: {filtered_cooccurrence_df.select(col("diseaseFromSourceMappedId"), col("targetFromSourceId")).dropDuplicates().count()}')
```
tskir, line 125: Also some literals:

```suggestion
    logging.info(f'Number of publications: {filtered_cooccurrence_df.select(col('tmp')).distinct().count()}')
    logging.info(f'Number of targets: {filtered_cooccurrence_df.select(col('targetFromSourceId')).distinct().count()}')
    logging.info(f'Number of diseases: {filtered_cooccurrence_df.select(col('diseaseFromSourceMappedId')).distinct().count()}')
    logging.info(f'Number of associations: {filtered_cooccurrence_df.select(col('diseaseFromSourceMappedId'), col('targetFromSourceId')).dropDuplicates().count()}')
```

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+                    col("text"),
+                    col('start1').alias('tStart'),
+                    col("end1").alias('tEnd'),
+                    col('start2').alias('dStart'),
+                    col("end2").alias('dEnd'),
+                    col('section')
+                )
+            ).alias('textMiningSentences'),
+            sum(col('evidence_score')).alias('resourceScore')
+        )
```
tskir, line 148:
```suggestion
        # Aggregating data by publication, target and disease:
        .groupBy(['pmid', 'targetFromSourceId', 'diseaseFromSourceMappedId'])
        .agg(
            first(col('targetFromSource')).alias('targetFromSource'),
            first(col('diseaseFromSource')).alias('diseaseFromSource'),
            collect_set(col('tmp')).alias('literature'),
            collect_set(
                struct(  
                    col('text'),
                    col('start1').alias('tStart'),
                    col('end1').alias('tEnd'),
                    col('start2').alias('dStart'),
                    col('end2').alias('dEnd'),
                    col('section')
                )
            ).alias('textMiningSentences'),
            sum(col('evidence_score')).alias('resourceScore')
        )
```

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+
+        # Summarizing all scores then filter for evidence string with at least score == 2
+        # .withColumn('resourceScore', sum(col('evidence_score')).over(w))
+        # .withColumn('resourceScores', collect_list(col('evidence_score')).over(w))
```
tskir, line 152: Ah, I suppose this relates to the commented-out code above which defined the `w`. I suggest:
* If this is never required, simply remove it;
* If this is sometimes required, make it an argument parser option and invoke only if needed

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+    (
+        aggregated_df
+
+        # Adding literal columns:
+        .withColumn('datasourceId',lit('europepmc'))
+        .withColumn('datatypeId',lit('literature'))
+
+        # Reorder columns:
+        .select(["datasourceId", "datatypeId", "targetFromSource", "targetFromSourceId", 'resourceScore',
+                "diseaseFromSource","diseaseFromSourceMappedId","literature","textMiningSentences"])
```
tskir, line 169: Final place with the literal mismatch :)

```suggestion
        # Reorder columns:
        .select(['datasourceId', 'datatypeId', 'targetFromSource', 'targetFromSourceId', 'resourceScore',
                 'diseaseFromSource', 'diseaseFromSourceMappedId', 'literature', 'textMiningSentences'])
```

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+
+        # Save output:
+        .write.format('json').mode('overwrite').option('compression', 'gzip').save(args.outputFile)
+    )
+    
+    logging.info(f'EPMC disease target evidence saved.')
+
+    return 0
```
tskir, line 177: Since this value isn't checked anyway, I think we can safely remove this

```suggestion
```

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+if __name__ == '__main__':
+
+    main()
```
tskir, line 182:
```suggestion
if __name__ == '__main__':
    main()
```

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+    ##
+    ## Load/filter datasets
+    ##
+    # partitionKeys = ['pmid', 'targetFromSourceId', 'diseaseFromSourceMappedId']
+    # w = Window.partitionBy(*partitionKeys)
+    filtered_cooccurrence_df = (
+        # Reading file:
+        spark.read.parquet(cooccurrenceFile)
+
+        # Filtering for diases/target cooccurrences:
```
tskir, line 104:
```suggestion
        # Filtering for disease/target cooccurrences:
```

modules/EPMC.py
```diff
@@ -0,0 +1,183 @@
...
+    # Report number of evidence:
+    logging.info(f'Number of evidence: {aggregated_df.count()}')
+
+    # Final formatting and saving data:
+    (
+        aggregated_df
+
+        # Adding literal columns:
+        .withColumn('datasourceId',lit('europepmc'))
+        .withColumn('datatypeId',lit('literature'))
```
tskir, line 165:
```suggestion
        # Adding literal columns:
        .withColumn('datasourceId', lit('europepmc'))
        .withColumn('datatypeId', lit('literature'))
```

tskir (changes requested):
Going forward, what is going to be our approach regarding the documentation? For example, this parser has a few notes (both in the PR description and in the code itself) which it would be nice to put somewhere. Perhaps README, until we decide on how to restructure the whole thing?

modules/EPMC.py
DSuveges: That's a really good point. I was also thinking about it earlier that we are kind of defying the whole point of the `if __name__ == "__main__":` condition. Although so far it is only your parser that follows this pattern, we should move into this direction.

DSuveges: Also if we keep "importability" in mind, the initialization of the logger should also be in the `if` block, and potentially the initiation of the spark session as well. We discuss it later.

modules/EPMC.py
DSuveges: I don't think so. These are mostly remarks for myself, I should have removed it. Regarding documentation in general, I would not go excessive with it. The readme of the repository should provide enough information to set up the environment and run the pipeline. The individual script should provide enough help on how it should be called (`--help`), that's probably the easiest thing to keep up to date. Also a good logging is informative. My biggest concern about documentation is that there's a big risk that the documentation, wherever it lives, diverge from the actual behaviour. 

Other than there shouldn't be really a need for anything else.

modules/EPMC.py
DSuveges: The threshold is quite arbitrary. The distribution of the length of the sentences is extremely skewed to the left: the median length of the 61M sentences is 230, while the mean length is 900. This extreme skewness is mostly due to problems in splitting the text into sentences. 

I picked 600 as a kind of good-enough threshold where (based on a random sample) 15% of the sentences were genuine sentences while the rest were improperly split sentences or concatenated lists from the paper. By excluding sentences above 600 we are losing ~8000 sentences representing 0.01% of all sentences.

modules/EPMC.py
DSuveges: It won't work in f-stings.  It needs to be resolved by this:
```python
logging.info(f'Number of publications: {filtered_cooccurrence_df.select(col("tmp")).distinct().count()}')
```
to this:

```python
logging.info(f"Number of publications: {filtered_cooccurrence_df.select(col('tmp')).distinct().count()}")
```

modules/EPMC.py
DSuveges: In an earlier version of the script we used windowing, but when I tried to use windows to summarize the score, it was failing (hence the comment in the beginning of the script about the memory issue)

modules/EPMC.py
DSuveges: Actually the danger has already kicked in: `sum` imported from `pyspark.sql.functions` overwrites the base sum function.

modules/EPMC.py
tskir: Agreed. Especially about bundling the documentation with the codebase (in the form of parser help/comments/docstrings), this is indeed a great way to ensure it doesn't get out of sync with reality

modules/EPMC.py
tskir: > It won't work in f-stings.

Indeed. In this case this wasn't a particularly well thought through suggestion from my side. Apologies and thanks for noticing :)

modules/EPMC.py
tskir: Thank you for the explanation! Do you think we could add it to the script, perhaps as a comment in this particular place where `600` first appears?

modules/EPMC.py
```diff
@@ -0,0 +1,179 @@
...
+import argparse
+import sys
+import logging
+
+import pyspark
+from pyspark.sql.types import StringType
+from pyspark.sql import SparkSession 
+import pyspark.sql.functions as pf
+from pyspark import SparkContext
+from pyspark.conf import SparkConf
```
tskir, line 12: Alphabetical order as per PEP8

```suggestion
import argparse
import logging
import sys

import pyspark
from pyspark import SparkContext
from pyspark.conf import SparkConf
from pyspark.sql import SparkSession 
from pyspark.sql.types import StringType
import pyspark.sql.functions as pf
```

modules/EPMC.py
```diff
@@ -0,0 +1,179 @@
...
+
+# The following target labels are excluded as they were grounded to too many target Ids
+EXCLUDED_TARGET_TERMS = ['TEC', 'TECS', 'Tec', 'tec', '\'', '(', ')', '-', '-S', 'S', 'S-', 'SS', 'SSS',
+         'Ss', 'Ss-', 's', 's-', 'ss', 'U3', 'U6', 'u6', 'SNORA70', 'U2', 'U8']
+
+def main(cooccurrenceFile, outputFile, local=False):
+
+    ##
+    ## Initialize spark session
+    ##
```
tskir, line 23: PEP8 says [block comments](https://www.python.org/dev/peps/pep-0008/#block-comments) should start with a single '#' followed by a single space:

```suggestion
    # Initialize spark session:
```

modules/EPMC.py
```diff
@@ -0,0 +1,179 @@
...
+            SparkSession.builder
+                .config(conf=sparkConf)
+                .getOrCreate()
+        )
+
+    logging.info(f'Spark version: {spark.version}')
+
+    ##
+    ## Log parameters:
+    ##
```
tskir, line 54:
```suggestion
    # Log parameters:
```

modules/EPMC.py
```diff
@@ -0,0 +1,179 @@
...
+    logging.info(f'Cooccurrence file: {cooccurrenceFile}')
+    logging.info(f'Output file: {outputFile}')
+    logging.info(f'Generating evidence:')
+
+    ##
+    ## Load/filter datasets:
+    ##
```
tskir, line 61:
```suggestion
    # Load/filter datasets.
```

modules/EPMC.py
```diff
@@ -0,0 +1,179 @@
...
+        # Reading file:
+        spark.read.parquet(cooccurrenceFile)
+
+        # Filtering for disease/target cooccurrences:
+        .filter(
+            (pf.col('type') == 'GP-DS') &  # Filter gene/protein - disease cooccurrence
+            (pf.col('isMapped') == True) &  # Filtering for mapped cooccurrences
+            (pf.col('pmid').isNotNull()) &  # Excluding publications without pmid
+            (pf.length(pf.col('text')) < 600) & # Exclude sentences with more than 600 characters
+            (pf.col('label1').isin(EXCLUDED_TARGET_TERMS) == False) # Excluding target labels from the exclusion list
```
tskir, line 72: PEP8 strikes again; [it wants at least two spaces](https://www.python.org/dev/peps/pep-0008/#inline-comments) between a statement and a comment

```suggestion
            (pf.length(pf.col('text')) < 600) &  # Exclude sentences with more than 600 characters
            (pf.col('label1').isin(EXCLUDED_TARGET_TERMS) == False)  # Excluding target labels from the exclusion list
```

modules/EPMC.py
```diff
@@ -0,0 +1,179 @@
...
+        # Summarizing all scores then filter for evidence string with at least score == 2
+        .filter(pf.col('resourceScore') > 1)
+    )
+
+    # Report number of evidence:
+    logging.debug(f'Number of evidence: {aggregated_df.count()}')
+
+    ##
+    ## Final formatting and saving data:
+    ##
```
tskir, line 121:
```suggestion
    # Final formatting and saving data:
```

modules/EPMC.py
```diff
@@ -0,0 +1,179 @@
...
+    cooccurrenceFile = args.cooccurrenceFile
+    logFile = args.logFile
+    outputFile = args.outputFile
+    local = args.local
+
+    return (cooccurrenceFile, logFile, outputFile, local)
+
+
+
+if __name__ == '__main__':
```
tskir, line 159: PEP8 again; exactly two blank lines between top level statements

```suggestion
    return (cooccurrenceFile, logFile, outputFile, local)


if __name__ == '__main__':
```

modules/EPMC.py
```diff
@@ -0,0 +1,179 @@
...
+    
+    # Parse arguments:
+    (cooccurrenceFile, logFile, outputFile, local) = parse_args()
```
tskir, line 162:
```suggestion
    cooccurrenceFile, logFile, outputFile, local = parse_args()
```

modules/EPMC.py
```diff
@@ -0,0 +1,179 @@
...
+        format='%(asctime)s %(levelname)s %(module)s - %(funcName)s: %(message)s',
+        datefmt='%Y-%m-%d %H:%M:%S',
+    )
+    if logFile:
+        logging.config.fileConfig(filename=logFile)
+    else:
+        logging.StreamHandler(sys.stderr)
+
+
+    # Calling main function:
```
tskir, line 177: And again as of PEP8, just one blank line _within_ the code block

```suggestion
        logging.StreamHandler(sys.stderr)

    # Calling main function:
```

modules/EPMC.py
```diff
@@ -0,0 +1,179 @@
...
+                    pf.col('start2').alias('dStart'),
+                    pf.col('end2').alias('dEnd'),
+                    pf.col('section')
+                )
+            ).alias('textMiningSentences'),
+            pf.sum(pf.col('evidence_score')).alias('resourceScore')
+        )
+
+        # Summarizing all scores then filter for evidence string with at least score == 2
+        .filter(pf.col('resourceScore') > 1)
```
tskir, line 113: The comment implies filtering on `>= 2`, but the actual filter is `> 1`. Granted, if those scores are always integers, then it's the same thing, but I think it's still better if the semantics is the same in the code and in the comment.

Also, I'm a bit confused, isn't the score supposed to be normalised to [0, 1] already? Or is this something done by the ETL now?

modules/EPMC.py
```diff
@@ -0,0 +1,179 @@
...
+from pyspark.sql.types import StringType
+from pyspark.sql import SparkSession 
+import pyspark.sql.functions as pf
+from pyspark import SparkContext
+from pyspark.conf import SparkConf
+
+
+# The following target labels are excluded as they were grounded to too many target Ids
+EXCLUDED_TARGET_TERMS = ['TEC', 'TECS', 'Tec', 'tec', '\'', '(', ')', '-', '-S', 'S', 'S-', 'SS', 'SSS',
+         'Ss', 'Ss-', 's', 's-', 'ss', 'U3', 'U6', 'u6', 'SNORA70', 'U2', 'U8']
```
tskir, line 17: Is this list ever likely to change? Also, is this something we receive from Miguel or could it be detected on the fly when ingesting the dataset?

modules/EPMC.py
```diff
@@ -0,0 +1,179 @@
...
+            (pf.length(pf.col('text')) < 600) & # Exclude sentences with more than 600 characters
+            (pf.col('label1').isin(EXCLUDED_TARGET_TERMS) == False) # Excluding target labels from the exclusion list
+        )
+        # Renaming columns:
+        .withColumnRenamed('keywordId1', 'targetFromSourceId')
+        .withColumnRenamed('keywordId2', 'diseaseFromSourceMappedId')
+        .withColumnRenamed('label1', 'targetFromSource')
+        .withColumnRenamed('label2', 'diseaseFromSource')
+
+        .withColumn('tmp', pf.col('pmid').cast(StringType()))
```
tskir, line 80: Suggest renaming `tmp` to something more informative like `pmid_str` or `pmid_str_tmp`

tskir (dismissed):
Like this a lot!

Added some more PEP8 nitpicking (luckily it is all in the form of suggestions and can be applied directly from GitHub) and a few final questions

modules/EPMC.py
DSuveges: The raw scores for a single cooccurrence range between 1-10, these scores for all cooccurrences then summarized for one publication to get the `associationScore`. This is what the parser generates for the ETL. 

This score is then divided by 100 and capped at 1. All within ETL.

modules/EPMC.py
DSuveges: I consider it as an interim solution. We are having some plans to improve grounding, once those improvements are in place, having this list and the filter in place we won't need that. 

This list was created based on my analysis showing these labels were notorious for being mapped to too many target IDs. But again, this issue should be treated systematically at the level of grounding. David/Miguel are aware of the issue, but due to time constraints, we could not address the issue.

ireneisdoomed (approved):
PEPfect 🙃

PR #84, opened 2021-06-07
Evidence generation without EFO

DSuveges:
- [ ] This PR updates parsers so evidence without mapped EFO terms got also saved. 
- [ ] Extra logic from certain parsers were removed to save unmapped diseases.
- [ ] The necessary updates in the snakefile was also made.


modules/SLAPEnrich.py
```diff
@@ -27,53 +30,51 @@ def generateEvidenceFromSource(self, inputFile, diseaseMapping, skipMapping):
...
+            .read.csv(diseaseMapping, sep=r'\t', header=True)
+            .select('Cancer_type_acronym', 'EFO_id')
+        )
 
         self.dataframe = self.dataframe.join(
             diseaseMappingsFile,
-            on="Cancer_type_acronym",
-            how="inner"
+            on='Cancer_type_acronym',
+            how='right'
```
DSuveges, line 77: This should be left.

Snakefile
```diff
@@ -22,23 +22,29 @@ rule all:
         GS.remote(f"{config['Gene2Phenotype']['outputBucket']}/gene2phenotype-{timeStamp}.json.gz"),
         GS.remote(f"{config['CRISPR']['outputBucket']}/crispr-{timeStamp}.json.gz"),
         GS.remote(f"{config['SysBio']['outputBucket']}/sysbio-{timeStamp}.json.gz"),
+        # GS.remote(f"{config['Phenodigm']['outputBucket']}/phenodigm-{timeStamp}.json.gz"),
+        # directory(GS.remote(f"{config['EPMC']['outputBucket']}/epmc-{timeStamp}")),
+        # directory(GS.remote(gs://genetics-portal-analysis/l2g-platform-export/data/genetics_portal_evidence.json.gz)),
```
tskir, line 27: Could you comment a bit on these lines and why they're commented out? Are we planning to add them in the future?

Also, the PhenoDigm line seems to occur twice (one instances is commented out and another one is not)

Snakefile
```diff
@@ -51,8 +57,7 @@ rule clingen:
         """
         python modules/ClinGen.py \
         --input_file {input} \
-        --output_file {output.evidenceFile} \
-        --unmapped_diseases_file {output.unmappedDiseases}
```
tskir, line 55: I'm not sure we necessarily want to remove this output: unmapped diseases can be useful for informing the manual curation efforts.

However, it's not a problem for now. I can reimplement output of the unmapped diseases in a consistent way for all parsers at some point in the future.

modules/CRISPR.py
```diff
@@ -73,56 +36,50 @@ def main():
...
+            {'Name': lambda x: list(x)}
+        )
+        .reset_index()
+    )
 
     # Updating columns:
     pooled_annotation = (
         pooled_annotation
-        .drop(['method',],axis=1)
+        .drop(['method', ], axis=1)
```
tskir, line 59: For list literals of length 1 the trailing comma is not required (it is required for tuples, though)
```suggestion
        .drop(['method'], axis=1)
```

modules/ClinGen.py
```diff
@@ -79,56 +67,47 @@ def generate_evidence_strings(self, filename):
...
+            # Generating evidence for all mapped efo:
+            if efo_mappings:
+                for efo_mapping in efo_mappings:
+                    evidence_with_efo = evidence.copy()
+                    evidence_with_efo['diseaseFromSourceMappedId'] = ontoma.interface.make_uri(efo_mapping['id']).split('/')[-1]
+                    self.evidence_strings.append(evidence_with_efo)
+                    if disease_name == 'adenosine kinase deficiency':
+                        print(evidence_with_efo)
+            else:
+                print('pocok')
```
tskir, line 88: Looks like a debug print, can probably be removed?

modules/ClinGen.py
```diff
@@ -79,56 +67,47 @@ def generate_evidence_strings(self, filename):
...
-                }
 
+            # Generating evidence for all mapped efo:
+            if efo_mappings:
+                for efo_mapping in efo_mappings:
+                    evidence_with_efo = evidence.copy()
+                    evidence_with_efo['diseaseFromSourceMappedId'] = ontoma.interface.make_uri(efo_mapping['id']).split('/')[-1]
+                    self.evidence_strings.append(evidence_with_efo)
+                    if disease_name == 'adenosine kinase deficiency':
+                        print(evidence_with_efo)
```
tskir, line 86: Also looks like a debug print

tskir (changes requested):
Everything looks good to me aside from a few really minor comments. Thank you for all of the style improvements!

Snakefile
DSuveges: So all the green rows are added in the PR, however I'm not sure why the ones commented are remained commented. Fix them.

DSuveges (commented):
Please let me know if you have found any other issues. Thank for reviewing all these changes.

Snakefile
tskir: Ah, of course, makes sense! I do feel silly now :)

PR #86, opened 2021-07-06
Bug fix: Cast resourceScore as float

tskir:
Fixing that annoying bug where Spark would coerce everything into float for internal comparisons, but still output the score as string and break the ETL pipeline

(Resubmitted with the proper branch name)


tskir:
@DSuveges Can you approve this change please so that I could merge it?

DSuveges (approved):
All looks good.

PR #87, opened 2021-07-12
Set up a reproducible Conda pipeline approach

tskir:
Closes https://github.com/opentargets/platform/issues/1628.

Following the advice outlined in [“Reproducible and upgradable Conda environments”,](https://pythonspeed.com/articles/conda-dependency-management/) this PR creates two distinct files: one is maintained manually and contains only the direct dependencies, while the second one is generated from it to provide a completely reproducible snapshot of the repository's environment at certain point in history.

Also added instructions on maintaining the new approach and removed the obsolete environment configuration files.


DSuveges (approved):
All looks good. The approach really makes sense.

PR #88, opened 2021-07-15
PanelApp rewrite

tskir:
Points marked with checkboxes are discussions which need to be resolved before merging.

### Data handling changes
* [x] Some phenotype names start with a question mark, for example: `Keratitis 148190 AD;Foveal hypoplasia 1 136520 AD;?Coloboma, ocular 120200 AD`. This is now treated as doubt about the accuracy, and such records are removed.
* [x] The `diseaseFromSourceId` use OMIM, if present, and any other ontology otherwise. This isn't relevant to this specific data source, but perhaps the fact that this field only allows for one ontology is worth revisiting in general.
* [x] Previously, in case the phenotype was absent, panel name was used instead. However, panel names are not really phenotypes, they represent high level grouping of phenotypes. Hence, this logic is not replicated in the rewrite. Instead, the records which have neither a phenotype nor an ontology code are removed as fundamentally unmappable.
* [x] Do we want the `diseaseFromSourceId` and `diseaseFromSource` contain the completely raw data, or the cleaned up versions?

### Parsing improvements
* Clean phenotype strings before splitting by semicolon and fix some common issues.
* Clean phenotype strings after splitting to remove extraneous information.
* Extract ontologies other than OMIM.
* Increase OMIM parsing accuracy.
* Increase publication ID parsing accuracy from API.

### Technical changes
* Reimplemented the module in PySpark.
* Updated PySpark dependency (2.4.0 → 3.1.2) and re-generated the locked environment file.
* Removed the old module `GenomicsEnglandPanelApp.py` and the partial refactor `PanelAppRevamp.py`. The new module is called simply `PanelApp.py`.


modules/PanelApp.py
```diff
@@ -0,0 +1,227 @@
...
+            os.rename(os.path.join(tmp_dir_name, json_chunks[0]), output_file)
+
+    def fetch_literature_references(self, all_panel_ids):
+        """Queries the PanelApp API to extract all literature references for (panel ID, gene symbol) combinations."""
+        publications = []  # Contains tuples of (panel ID, gene symbol, PubMed ID).
+        for panel_id in all_panel_ids:
+            logging.debug(f'Fetching literature references for panel {panel_id!r}.')
+            url = f'https://panelapp.genomicsengland.co.uk/api/v1/panels/{panel_id}'
+            for gene in requests.get(url).json()['genes']:
+                for publication_string in gene['publications']:
```
DSuveges, line 190: Something strange is happening here. Many of the genes have pubmed id `2`. Which is most likely an artefact. 

Let's say panel id: `234`, this function returns a dataframe in which the following rows exist:

```
+--------+------+-------------+
|Panel ID|Symbol|   literature|
+--------+------+-------------+
|     234| ACTG2|          [2]|
|     234|TBC1D1|          [2]|
|     234| GRIP1|[2, 24700879]|
|     234| SALL4|          [2]|
|     234| TRAP1|          [2]|
|     234| FREM1|          [2]|
+--------+------+-------------+
```
This is not good. If I take a look at the API response I see:
```
ACTG2   - ['PMID: 25998219']
FREM1   - ['PMID: 24700879']
GRIP1   - ['PMID: 22510445', '24700879', '14730302.']
TBC1D1  - ['PMID: 26572137']
TRAP1   - ['PMID: 24152966 - recessive mutations reported in 2 families with CAKUT, and 3 families with VACTERL.']
SALL4   - ['PMID: 20301547']
```
Most likely there is some issue with the regular expression cleaning these strings. However I found it strange that GRIP1 has only 2 pmids, indicating more than one were resolved to `2` then cased into a set.

modules/PanelApp.py
```diff
@@ -0,0 +1,227 @@
...
+    PMID_RE = [
+        (
+            r'^'                # Start of the string.
+            r'[\d, ]+'          # A sequence of digits, commas and spaces.
+            r'(?: |$)'          # Ending either with a space, or with the end of the string.
+        ),
+        (
+            r'(?:PubMed|PMID)'  # PubMed or a PMID prefix.
+            r'[: ]*'            # An optional separator (spaces/colons).
+            r'[\d, ]'           # A sequence of digits, commas and spaces.
```
DSuveges, line 65:
```suggestion
            r'[\d, ]+'           # A sequence of digits, commas and spaces.
```
I think the number of digits should be more than one, as it causes side effects. See my other comment on row 190. (I think it would be a good idea to show one-one typical examples the patterns are matching)

modules/PanelApp.py
```diff
@@ -0,0 +1,227 @@
...
+            .withColumnRenamed('phenotype', 'diseaseFromSource')
+            # diseaseFromSourceId populated above
+            # literature populated above
+            .withColumnRenamed('Panel Id', 'studyId')
+            .withColumnRenamed('Panel Name', 'studyOverview')
+            .withColumnRenamed('Symbol', 'targetFromSourceId')
+        )
+
+        # Save data.
+        with tempfile.TemporaryDirectory() as tmp_dir_name:
```
DSuveges, line 174: I like this! Writing a temporary file, doing a check and then renaming. Once we are touching the other parsers, we can apply this practice.

DSuveges (changes requested):
I didn't know where to put this comment, but there are 54+k rows in the output, however there's a great redundancy: only 20k unique. I'm not sure at which point this redundancy is introduced, but we need to drop duplicates.

ireneisdoomed (commented):
I really love how all the transformations and cleaning steps are performed. A few comments on your notes:
1. The question mark is a very good point. As the strings were so dirty, I always assumed these were artifacts. I think it is worth asking Genomics England about this. I can do it.
2. I already commented that I'm not sure about the preference of OMIM over other ontologies (taking into account that OMIM is the most prevalent one)
3. Although you're right about the panel name not being a phenotype, it still is a group of diseases that are susceptible to being mapped to an EFO ID. Of course it is more desirable to have a more granular term, but I don't see the reason why mapping it to the panel name is wrong.
4. Reporting the raw string would imply to work with two versions: the one that we process and the one that is unaltered. I don't think we are introducing a substantial change to the original string other than cleaning it. I'm in favour of using the cleaned up version for the sake of simplicity.

tskir:
Thank you both for the comments!

@DSuveges Regarding the duplication issues, there turned out to be three sources:
1. We're only ingesting a subset of columns: `('Symbol', 'Panel Id', 'Panel Name', 'Mode of inheritance', 'Phenotypes')`. While the source data don't contain any duplicates, uniqueness is not guaranteed for that column subset. Some records only differ in other columns, like `Sources` or `Super Panel Id`. Resolved by applying `.distinct()` after column selection. After resolving this, **103** duplicates remained.
2. An additional pre-semicolon-splitting cleanup operation was previously overlooked: a semicolon can have extra leading or trailing spaces. This leads to cases like `Alport Syndrome` and ` Alport Syndrome` (with an extra space symbol) both being present in `cohortPhenotypes`, but after post-splitting cleanup evaluating to the same record in `diseaseFromSource`. Resolved by adding additonal pre-splitting cleanup operation. After resolving this, **25** duplicates remained.
3. The remaining duplicates are similar to case 2, but are not easily correctable without altering the semantics of the `cohortPhenotypes` representation, for example `Pontocerebellar hypoplasia type 2D (613811)` and `Pontocerebellar hypoplasia type 2D, 613811`. After resolving this, no duplicates remained, and the set now has 20,235 evidence strings in total.

The remaining comments will be addressed inline, I'm still finishing some work on them

modules/PanelApp.py
tskir: Thank you for noticing! It was indeed due to a regex issue. Now corrected, & also slightly improved the post-filtering of PMIDs: https://github.com/opentargets/evidence_datasource_parsers/pull/88/commits/ac83edb4f832f10552faaff10e96eb036e53f1a0

modules/PanelApp.py
tskir: Yes, this was exactly the problem. Fix committed in https://github.com/opentargets/evidence_datasource_parsers/commit/ac83edb4f832f10552faaff10e96eb036e53f1a0, and examples are added in https://github.com/opentargets/evidence_datasource_parsers/pull/88/commits/d49cc8440396258d271e63ca04d71f8bf211a661.

Also, all original + parsed PMIDs are available here: https://docs.google.com/spreadsheets/d/1otdZXXEmyNaLT7j67DwgHUM-pHPJaYc0SQnyqb1bUtA/edit?usp=sharing.

modules/PanelApp.py
tskir: The main utility of this code fragment is making sure that the output is indeed one single file, and this is what the majority of it (7 out of 8 lines) is doing.

By default, if we just call `.coalesce(1).write`, it will create a directory with one, randomly named, gzipped JSON inside, which is not always convenient. By doing the chunk discovery and renaming, we are making sure that there is only one file and it is consistently named.

The remaining 1 line, the assert, is just there as a precaution for a very unlikely scenario. It's incredibly cheap to write (a few seconds of my time), but can trigger in certain situations. For example, if we ever run the parsers on network filesystems (NFS, fuse mounts), it's a very useful check to have.

tskir:
@ireneisdoomed Addressing your comments:

> The question mark is a very good point. As the strings were so dirty, I always assumed these were artifacts. I think it is worth asking Genomics England about this. I can do it.

There's actually a remaining action in https://github.com/opentargets/platform/issues/1636 for me to compile all irregularities for reporting to Genomics England, so I think we can do it as a part of that effort.

> I already commented that I'm not sure about the preference of OMIM over other ontologies (taking into account that OMIM is the most prevalent one)

As I mentioned, in the current version there are no cases where both an OMIM and a non-OMIM ontology are present in the same record. (And if in the future such cases do occur, we don't have a way to represent them in the evidence strings anyway.) Since OMIM is used a de facto standard in the PanelApp, taking it by default is probably fine, at least for the time being.

> Although you're right about the panel name not being a phenotype, it still is a group of diseases that are susceptible to being mapped to an EFO ID. Of course it is more desirable to have a more granular term, but I don't see the reason why mapping it to the panel name is wrong.

I think you're right. I've taken a closer look at the panel names, and in fact some of them are not even groups of diseases, but single specific conditions. I will reimplement using a panel name in case no phenotypes are specified.

> Reporting the raw string would imply to work with two versions: the one that we process and the one that is unaltered. I don't think we are introducing a substantial change to the original string other than cleaning it. I'm in favour of using the cleaned up version for the sake of simplicity.

As discussed during today's stand-up meeting, I'll continue to use (mostly) the original names in `cohortPhenotypes`, with only very minor alterations like whitespace trimming; and `diseaseFromSource`/`diseaseFromSourceId` will contain the cleaned up/extracted versions.

I will also trace which entries from `cohortPhenotypes` get transformed into which `diseaseFromSource` records and remove any duplicates, as discussed.

modules/PanelApp.py
```diff
@@ -0,0 +1,227 @@
...
+    array, array_distinct, col, collect_set, concat, explode, lit, regexp_extract, regexp_replace, split, trim, when
+)
+
+
+class PanelAppEvidenceGenerator:
+
+    # Fixes which are applied to original PanelApp phenotype strings *before* splitting them by semicolon.
+    PHENOTYPE_BEFORE_SPLIT_RE = {
+        # Fixes for specific records.
+        r'\(HP:0006574;\);': r'(HP:0006574);',
```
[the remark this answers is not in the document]
tskir: Sometimes, the only way is to address the specific cases. For example, on the line highlighted in this thread, the source record looks like this: `(HP:0006574;);` — an extra semicolon interferes with phenotype splitting. Unfortunately there's no easy way to generalise this. I hope that once I report the list of issues to PanelApp, they can eventually clean this up and avoid similar things in the future, so that this becomes less reelvant.

We don't want to remove all round brackets in all cases, because sometimes they do convey meaningful grammar:
* `Peroxisome biogenesis disorder 7A (Zellweger)`
* `INFANTILE REFSUM DISEASE (IRD)`
* `Muscular dystrophy-dystroglycanopathy (congenital with mental retardation), type B, 6`

In contrast, the curly brackets are always removed, which is already implemented.

modules/PanelApp.py
```diff
@@ -0,0 +1,227 @@
...
+        # Fix cases like "Aarskog-Scott syndrome, 305400Mental retardation, X-linked syndromic 16, 305400", where
+        # several phenotypes are glued to each other due to formatting errors.
+        r'(\d{6})([A-Za-z])': r'$1;$2',
+
+        # Replace all tab/space sequences with a single space.
+        r'[\t ]+': r' ',
+    }
+
+    # Cleanup regular expressions which are applied to the phenotypes *after* splitting.
+    PHENOTYPE_AFTER_SPLIT_RE = (
```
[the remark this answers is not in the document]
tskir: Thank you so much for pointing this out! Thanks to this comment I was able to find yet another type casting related bug. Apparently, the `col('Panel Version') > 1` filter did not always perform correct comparisons. Fixed in https://github.com/opentargets/evidence_datasource_parsers/pull/88/commits/7a8910d88a837f6ed831d85ce161ba00cf6193ca

I'll now take a look at `NO OMIM NUMBER` and other ways in which this lack of information could be represented.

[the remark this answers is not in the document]
tskir: The regular expressions for cleaning up PMIDs and phenotypes have been significantly updated now, including the `NO OMIM NUMBER` case and many others. I'll post a separate update when I submit the PR for the next review round

modules/PanelApp.py
```diff
@@ -0,0 +1,227 @@
...
+                )
+            )
+            .withColumn('phenotype', regexp_replace(col('phenotype'), f'({self.OTHER_RE})', ''))
+
+            # Extract OMIM identifiers and remove them from the phenotype string.
+            .withColumn('omim_id', regexp_extract(col('phenotype'), self.OMIM_RE, 2))
+            .withColumn('omim', when(col('omim_id') != '', concat(lit('OMIM:'), col('omim_id'))))
+            .withColumn('phenotype', regexp_replace(col('phenotype'), f'({self.OMIM_RE})', ''))
+
+            # Choose one of the ontology identifiers, keeping OMIM as a priority.
```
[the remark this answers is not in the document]
tskir: As discussed elsewhere, this doesn't really have an effect in this case because OMIM and other ontologies don't co-occur in the same record. OMIM is the de factor standard for the PanelApp, so it seems to make sense to prefer it

tskir:
@DSuveges @ireneisdoomed Ready for the next round of reviews. Summary of all changes since the original submission:
- Fix evidence string duplication issues
- Fix PMID parsing issues
- Add examples of PMID matching patterns
- Fix filtering by panel version ≥ 1.0
- Implement outputting phenotype and PMID intermediate dataframes for debugging (highly useful now & in the future)
- Check and amend phenotype cleanup regular expressions
- Check and amend PMID cleanup regular expressions
- Reimplement using the panel names as fallback when any meaningful phenotypes are missing

I've also looked into ways of removing near-duplicate entries from `cohortPhenotypes` (the ones which map to the same `diseaseFromSource` + `diseaseFromSourceId` combination). I propose that we do _not_ implement this now for the following reasons:
- Because of the way the processing is currently organised in PySpark, it would be very complicated to implement, probably requiring a UDF to track all of the conversion steps.
- It only affects about a couple dozen records, so the effort isn't justified by the possible benefits.
- And besides, the nature of the `cohortPhenotypes` field is to contain the original list as specified by the data source. Since it's only used for provenance purposes, we probably shouldn't spend significant time on doing any special post-processing.

Also here are some resources for you convenience when doing the reviews:
* The latest evidence strings:
[panelapp.json.gz](https://github.com/opentargets/evidence_datasource_parsers/files/6890491/panelapp.json.gz). The number of the evidence strings is now 44774, all unique. The increase is mostly due to fixing the panel version filtering bug.
* Spreadsheet with the intermediate data to evaluate the quality of cleaning up phenotypes and PMIDs: https://docs.google.com/spreadsheets/d/1otdZXXEmyNaLT7j67DwgHUM-pHPJaYc0SQnyqb1bUtA/edit?usp=sharing

ireneisdoomed (commented):
I have run the script and the generated evidence strings look good and validate. @tskir, I think it would be informative to add some logging messages to track the parser run and some info about the output, number of evidence at least (plus distinct targets and diseases?).

On a separate note, I attempted to run it via snakemake and I got a SyntaxError in line 26. This line was uncommented by @DSuveges to include EPMC and genetics portal in the pipeline. Daniel, do you remember testing this code?
These two were added by me some time ago, if you're both OK I will open an issue to trace the bug and amend it.

DSuveges:
@ireneisdoomed, yes, please open a ticket to clean up the snakemake file.

modules/PanelApp.py
```diff
@@ -0,0 +1,321 @@
...
+    def __init__(self, debug_output_prefix=None):
+        self.spark = SparkSession.builder.appName('panelapp_parser').getOrCreate()
+        self.debug_output_phenotypes_filename = None
+        self.debug_output_pmids_filename = None
+        self.debug_output_pmids_data = set()
+        if debug_output_prefix:
+            self.debug_output_phenotypes_filename = f'{debug_output_prefix}.phenotypes.tsv'
+            self.debug_output_pmids_filename = f'{debug_output_prefix}.pmids.tsv'
+
+    def __del__(self):
```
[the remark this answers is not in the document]
tskir: In this particular case this is what happens:
* When the object is created (`__init__`), various parameters, including the debugging facilities, are set up
* Over the course of the object's lifetime, debugging information is added to `self.debug_output_pmids_data`
* Finally, when the object is destroyed (`__del__`), that information is saved to the debugging output file.

This approach allows the main code to focus on the actual data transformations, and lets Python's internal facilities to deal with the wrap-up and output of the debugging information separately

tskir:
@ireneisdoomed Thank you for the comments! I've addressed them now. And regarding the general points:

> I think it would be informative to add some logging messages to track the parser run

Indeed, I've now added this in https://github.com/opentargets/evidence_datasource_parsers/pull/88/commits/3d30fa2119a14826ab09505c3fd8bea64a9b7324.

> and some info about the output, number of evidence at least (plus distinct targets and diseases?)

That's a very good idea, although I think it makes sense to do this uniformly for all parsers. Added a point to this ticket: https://github.com/opentargets/platform/issues/1574.

Ready for the third iteration of reviews now :crossed_fingers:

ireneisdoomed (approved):
Thank you for the changes! For me it is good to go :)

PR #91, opened 2021-08-11
A parser is added to generate evidence for partner CRISPR screens

DSuveges:
The script reads a [study table](https://docs.google.com/spreadsheets/d/1eow_rTizGY_Vda0Q-8sFNWCOyFkKRSUen1Em1ysPSiA/edit#gid=0), which is a google spreadsheet saved as tsv.

It also expects to have a folder with the data files, which are expected to be separated for each OT project. Practically as downloaded from [here](https://drive.google.com/drive/u/1/folders/1WJ0MwWid2qDsADm-F0Pvp1mprQtEPs6M) .

The generated evidence looks like this:

```json
{
  "targetFromSourceId": "YIF1B",
  "diseaseFromSourceMappedId": "EFO_0000249",
  "projectId": "OTAR036",
  "studyId": "OTAR036_TAU_uptake_1",
  "studyOverview": "Comparison of functional populations of iPSC neurons based on their ability to take up monomeric/aggregated tau protein",
  "contrast": "TauNEGTransPOS (only transferrin taken up) vs TauPOSTransPOS  (both proteins taken up)",
  "crisprScreenLibrary": "Kosuke v1.1 (Behan et al., 2018)",
  "cellType": "iPSC derived cortical neurons",
  "cellLineBackground": "KOLF2-C1",
  "direction": "upper tail",
  "resourceScore": 0.0063227,
  "datasourceId": "ot_crispr",
  "datatypeId": "ot_partner"
}
```

The generated evidence is validating against the 2.0.10 schema.

To avoid conflicts, I have not updated the snakefile.


modules/OTAR_CRISPR.py
```diff
@@ -0,0 +1,181 @@
...
+                'targetFromSourceId', 'diseaseFromSourceMappedId',
+                'projectId', 'studyId', 'studyOverview', 'contrast', 'crisprScreenLibrary',
+                'cellType', 'cellLineBackground', 'geneticBackground',
+                'direction', 'resourceScore',
+                'datasourceId', 'datatypeId'
+            ]]
+        )
+
+    @staticmethod
+    def parse_MAGeCK_file(row) -> pd.DataFrame:
```
tskir, line 98:
```suggestion
    def parse_MAGeCK_file(row: pd.Series) -> pd.DataFrame:
```

modules/OTAR_CRISPR.py
```diff
@@ -0,0 +1,181 @@
...
+                pd.read_csv(datafile, sep='\t')
+                .rename(columns={filterColumn: 'resourceScore', 'id': 'targetFromSourceId'})
+                .loc[lambda df: df.resourceScore <= threshold]
+                [['targetFromSourceId', 'resourceScore']]
+                .assign(studyId=studyId)
+            )
+            logging.info(f'Number of genes reach threshold: {len(mageck_df)}')
+            return mageck_df
+        except FileNotFoundError:
+            logging.info(f'Study skipped as file was not found: {datafile}')
```
tskir, line 118: I would argue that this should be an unhandled, fatal exception situation rather than just skipping a study. A note in logs might be overlooked, but a crash probably not

modules/OTAR_CRISPR.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def get_evidence_data(self) -> pd.DataFrame:
+        """Returning the merged dataset"""
+        return self.merged_dataset
```
tskir, line 122: Unless I'm mistaken, this is not used anywhere?

modules/OTAR_CRISPR.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def write_evidence(self, output_file: str) -> None:
+        """Write the merged evidence to file"""
+
+        json_list = [json.dumps(row.dropna().to_dict()) for _, row in self.merged_dataset.iterrows()]
+        with gzip.open(output_file, 'wt') as f:
+            f.write('\n'.join(json_list))
```
tskir, line 129:
```suggestion
            f.write('\n'.join(json_list) + '\n')
```

To have that sweet final line break

tskir (commented):
I'm nitpicking here and there; but overall this is a super clean and readable parser, and all logic is easily traceable, I like it.

modules/OTAR_CRISPR.py
```diff
@@ -0,0 +1,181 @@
+
```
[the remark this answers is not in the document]
tskir: (Either that, or alternatively the first empty line could just be removed.)

modules/OTAR_CRISPR.py
```diff
@@ -0,0 +1,181 @@
...
+        self.study_file = study_file
+
+        # Read and process study table:
+        study_df = pd.read_csv(study_file, sep='\t')
+
+        # If present, the column description row is dropped:
+        if 'Studied cell type' in study_df.loc[0].tolist():
+            logging.info('Dropping column descriptions')
+            study_df.drop(0, axis=0, inplace=True)
```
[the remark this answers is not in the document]
DSuveges: Yes, that makes sense. I was also considering having the description mandatory, but I forgot that we can just skip that row from reading, so I always considered dropping. Your suggestion sounds much better.

modules/OTAR_CRISPR.py
```diff
@@ -0,0 +1,181 @@
...
+            # rename columns:
+            .rename(
+                columns={
+                    'isDerived': 'isCellTypeDerived',
+                    'library': 'crisprScreenLibrary',
+                    'mode': 'crisprStudyMode',
+                    'populations': 'contrast',
+                    'studyDescription': 'studyOverview'
+                }
+            )
```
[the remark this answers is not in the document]
DSuveges: Ys, makes sense.

modules/OTAR_CRISPR.py
DSuveges: Good point.

modules/OTAR_CRISPR.py
DSuveges: I actually used this method while I was developing. But yeah, not used in the production code.

modules/OTAR_CRISPR.py
```diff
@@ -0,0 +1,181 @@
...
+        # Concatenate all hits into one single dataframe:
+        hits_df = (
+            pd.concat(hits.to_list())
+            .reset_index(drop=True)
+        )
+
+        # Merging:
+        self.merged_dataset = (
+            self.study_df
+            .assign(direction=lambda df: df.filterColumn.map(FILTER_COLUMN_MAP))
```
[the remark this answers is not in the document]
DSuveges: Yes, that was an uncommitted change in the parser. I'm adding that update with the responses to your comments. The new field is `statisticalTestTail`

modules/OTAR_CRISPR.py
DSuveges: Especially true when using `logging.info`

modules/OTAR_CRISPR.py
```diff
@@ -0,0 +1,161 @@
...
+import argparse
+from datetime import datetime
+import gzip
+import json
+import logging
+import sys
+
+import pandas as pd
+
+# The statisticalTestTail is inferred by the column name wich is being filtered on:
```
ireneisdoomed, line 12:
```suggestion
# The statisticalTestTail is inferred by the column name which is being filtered on:
```

modules/OTAR_CRISPR.py
```diff
@@ -0,0 +1,161 @@
...
+        study_df = pd.read_csv(study_file, sep='\t', skiprows=[1])
+
+        self.study_df = (
+            study_df
+
+            # drop rows with no study id or data file:
+            .loc[study_df.studyId.notna() & study_df.dataFile.notna()]
+            .assign(
+                # Split mapped diseases to list:
+                diseaseFromSourceMappedId=study_df.diseases.str.replace(' ', '').str.split('|'),
```
ireneisdoomed, line 35: I am blown away by this use of `assign` and how it can replicate the `withColumn` function 😍
I have one question: when you refer to `study_df.diseases`, is this the whole initial column or the filtered one (after loc)?
In the Pandas documentation I see this could also be an alternative:
```
diseaseFromSourceMappedId=lambda X: X.diseases.str.replace(' ', '').str.split('|'),
```

modules/OTAR_CRISPR.py
```diff
@@ -0,0 +1,161 @@
...
+            [['targetFromSourceId', 'resourceScore']]
+            .assign(studyId=studyId)
+        )
+        logging.info(f'Number of genes reach threshold: {len(mageck_df)}')
+        return mageck_df
+
+    def write_evidence(self, output_file: str) -> None:
+        """Write the merged evidence to file"""
+
+        json_list = [json.dumps(row.dropna().to_dict()) for _, row in self.merged_dataset.iterrows()]
```
ireneisdoomed, line 107: to_json() supports exporting to JSON lines too. Do you think the dataframe could also be exported like this?

```
self.merged_dataset.dropna().to_json(output_file, compression='gzip', orient='records', lines=True)
```

ireneisdoomed (commented):
I totally second @tskir comment on readability. It is a joy to review.

I have a suggestion on reading the study table. I expect this table to change over time, so instead of downloading the file you can read the spreadsheet like this:
```

# arg.study_url = 'https://docs.google.com/spreadsheets/d/1eow_rTizGY_Vda0Q-8sFNWCOyFkKRSUen1Em1ysPSiA/edit?usp=sharing'
study_url = arg.study_url.replace('/edit?usp=sharing', '/export?format=csv')
study_df = pd.read_csv(study_url, skiprows=[1])

```

modules/OTAR_CRISPR.py
DSuveges: Yes. That's correct, however it's resolved by using the available index. You can see how it works in this example:

```python
import pandas as pd

df = pd.DataFrame([
    {'a': 2, 'b': 24},
    {'a': 4, 'b': None},
    {'a': 12, 'b': 100}
])

(
    df
    .loc[df.b.notna()]
    .assign(
        double=df.a*2
    )
)
```

modules/OTAR_CRISPR.py
DSuveges: No, unfortunately that is not good. It has unwanted side effects:

```python
df = pd.DataFrame([
    {'a': 2, 'b': 24},
    {'a': 4, 'b': None},
    {'a': 12, 'b': 100}
])

(
    df
    .dropna().to_json('test.json.gz', compression='gzip', orient='records', lines=True)

)
```

Gives this:
```
{"a":2,"b":24.0}
{"a":12,"b":100.0}
```

modules/OTAR_CRISPR.py
DSuveges: But in general I try to follow the canonical method using the `lambda`.

modules/OTAR_CRISPR.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+
+        # Concatenate all hits into one single dataframe:
+        hits_df = (
+            pd.concat(hits.to_list())
+            .reset_index(drop=True)
+        )
+
+        # Merging:
+        self.merged_dataset = (
```
ireneisdoomed, line 66: I didn't notice `self.merged_dataset` was not in the constructor. I tend to include all instance variables in `__init__`:
```
self.merged_dataset = None
```

DSuveges:
Given the exploratory nature of the partner preview instance datasources, these parsers will not yet be merged into the master branch. However will keep the `ds_1597_crispr_screen` branch.

PR #95, opened 2021-09-08
Deprecate gene mapping from PheWAS Catalog

ireneisdoomed:
PheWAS Catalog required gene symbol to ENSID mapping to merge it with the dataset that contains the variant information (`phewas_w_consequences.tsv`). There was a bug in this process, where the gene parser was returning empty strings.

This PR solves this data issue, addressed in [#1494](https://github.com/opentargets/platform/issues/1494) and [#1731](https://github.com/opentargets/platform/issues/1731).

- [X] I have regenerated the file we extract from the Variant Index to index the dataset using gene symbols, not ENSIDs.
- [X] I have removed the use of the GeneParser in the PheWAS script: `targetFromSource` is now null and `targetFromSourceId` contains Gene Symbols that will be handled by the ETL.
- [X] I have deleted `common/HGNCParser.py`

Since no gene mapping is happening on our side now, the number of evidence strings have increased. _We should be expecting an increase in the number of invalid evidence due to target._

To confirm I'm not introducing any bugs in the parser, I compared the associations with the old evidence file and extracted which were the offending targets:
```
+------------------+-----+
|targetFromSourceId|count|
+------------------+-----+
|        intergenic| 1101|
|        Intergenic|  506|
|             PCDP1|  133|
|      LOC100287944|  126|
|        HLA region|  121|
|         LOC643308|  116|
|         LOC441167|  109|
|      LOC100289473|   95|
|        HLA-region|   94|
|          MGC45800|   91|
|              FONG|   90|
|              PAPL|   87|
|             MYP11|   83|
|          FLJ41733|   80|
|      LOC100505844|   77|
|         LOC650157|   76|
|         LOC391642|   74|
|          MGC87042|   74|
|         C14orf181|   72|
|         LOC199897|   72|
|          FLJ32955|   72|
|         LOC285696|   70|
|              CHLI|   68|
|      LOC100132612|   67|
|          MGC34034|   64|
|      LOC100289058|   64|
|         LOC133789|   62|
|      LOC100652824|   62|
|      LOC100131818|   60|
|      LOC100132354|   59|
|         LOC284998|   58|
|         LOC387933|   49|
|          LOC63930|   49|
+------------------+-----+
```

Intergenic regions account for many new evidence strings. As discussed with @DSuveges, we will be dropping those as they don't fit in the concept of a target.

- [X] Added a filter on the gene column to remove DNA regions

That leaves us with **192477 evidence strings** (vs. 189971) that have been validated and uploaded to the corresponding bucket.

### `phewas_w_consequences.csv` generation

The method to generate this dataset is described in the relevant tickets: [#1243](https://github.com/opentargets/platform/issues/1243) and [#1494](https://github.com/opentargets/platform/issues/1494).

I'm sharing here the script used to generate this file, which we should be re running whenever the variant index changes.

https://gist.github.com/ireneisdoomed/635cb1c6327f9d2f0a05fc561d3cce2c


modules/PheWAS.py
```diff
@@ -232,7 +217,6 @@ def main(genesSet, inputFile, consequencesFile, diseaseMapping, skipMapping):
     parser.add_argument('-i', '--inputFile', required=True, type=str, help='Input .csv file with the table containing association details.')
     parser.add_argument('-c', '--consequencesFile', required=True, type=str, help='Input look-up table containing the variation consequences coming from the Variant Index.')
     parser.add_argument('-d', '--diseaseMapping', required=False, type=str, help='Input look-up table containing the phenotype mappings to an EFO ID.')
-    parser.add_argument('-g', '--genesSet', required=False, type=str, help='URL for the complete HGNC approved dataset in JSON format.')
     parser.add_argument('-o', '--outputFile', required=True, type=str, help='Name of the compressed json.gz output file containing the evidence strings.')
     parser.add_argument('-s', '--skipMapping', required=False, action='store_true', help='State whether to skip the disease to EFO mapping step.')
```
DSuveges, line 221: This is quite marginally related: if we want to skip the disease mapping we can just simply not pass the diseaseMaping file. The presence or absence can be used as a flag to indicate weather we want or not want disease mapping. But this is not a blocker for sure.

DSuveges (approved):
The modifications look good and reasonable. The updated script runs and generate evidence as expected.

modules/PheWAS.py
ireneisdoomed: Thanks for the comment. This was implemented when the plan was to modularise the disease mapping process and maybe move it to the ETL.

I'll bear this comment in mind and will change all the parsers accordingly.

PR #96, opened 2021-09-09
Update disease mapping process

ireneisdoomed:
This PR includes all the necessary changes to generate evidence strings for those parsers affected by the OnToma update (except for Phenodigm, for which Kirill is on top of):
- PanelApp.py
- Gene2Phenotype.py
- Orphanet.py
- Clingen.py

## General changes

- [X] `common.ontology.add_efo_mapping` is instantiated to populate the `diseaseFromSourceMappedId` field across all the mentioned parsers in a consistent way.
- [X] Several style changes to the parsers have been made so that we move towards the modularisation of the evidence generation. The general approach is to handle the 1) reading files, 2) processing into evidence format, 3) write evidence strings processes in separate functions that are invoked in `main`. 
- [X] Added  `cache_dir`  as an argument to store the OnToma input files.
- [X] Added `local` as an argument to specify the better Spark configuration

### Clingen.py

- [X] The whole script has been moved from Pandas into PySpark following the same style lines as the rest of the parsers. 

### Orphanet.py

- [X] Added filter to avoid having evidence strings with `targetFromSourceId == Null`
- [X] Removed duplicated PubMed ID's from the literature array
- [X] Removed whitespaces from PubMed ID's

### Gene2Phenotype.py
- [X] Improved parsing of diseaseFromSourceId. From 615860 to OMIM:615860 and also deleted the field when the value was `No disease mim`

Some columns of the input files were renamed and that required some changes (still not the data changes we are expecting from them):
- [X] `allelic requirement list` changed to `allelic requirement`. No more splitting is needed.
- [X] `phenotype list` is now `phenotypes`
- [X] `pmid list` is now `pmids`
- [X] `prev symbol list list` is now `prev symbols`
- [X] `gene disease entry date`is now `gene disease pair entry date`


modules/Gene2Phenotype.py
```diff
@@ -243,29 +159,29 @@ def map_disease(label, disease_id):
...
 
-    # Saving data:
-    (
-        evidence_df
-        .write.format('json').mode('overwrite').option('compression', 'gzip').save(outfile)
-    )
+def translate(mapping):
+    '''
+    Mapping consequences - to SO codes
+    '''
```
tskir, line 170:
```suggestion
    """
    Mapping consequences - to SO codes
    """
```

modules/Gene2Phenotype.py
```diff
@@ -243,29 +159,29 @@ def map_disease(label, disease_id):
...
+    def translate_(col):
+        return mapping.get(col)
+    return udf(translate_, StringType())
+
+def write_evidence_strings(evidence: DataFrame, output_file: str) -> None:
+    '''Exports the table to a compressed JSON file containing the evidence strings'''
```
tskir, line 176:
```suggestion
    """Exports the table to a compressed JSON file containing the evidence strings."""
```

tskir (commented):
Everything looks great, thank Irene!

I think we'll need to merge my PR first, because GitHub diffs are looking a bit weird (for example, it shows that `ontology.py` is newly added and there's no proper diff), and there may be some merge conflicts on top of it

ireneisdoomed:
The latest changes from Kirill's branch have been committed. This branch generates evidence properly and is ready for review.

modules/ClinGen.py
```diff
@@ -1,137 +1,163 @@
...
+
+    write_evidence_strings(evidence_df, output_file)
+    logging.info(f'{evidence_df.count()} evidence strings have been saved to {output_file}')
+
+def read_input_file(input_file: str, spark_instance) -> DataFrame:
+    '''
+    Reads Gene Validity Curations CSV file into a Spark DataFrame forcing the schema
+    The first 6 rows of this file include metadata that needs to be dropped
+    '''
+
```
tskir, line 66: Docstring as per PEP8

```suggestion
    """
    Reads Gene Validity Curations CSV file into a Spark DataFrame forcing the schema
    The first 6 rows of this file include metadata that needs to be dropped
    """

```

modules/ClinGen.py
```diff
@@ -1,137 +1,163 @@
...
+            col('DISEASE ID (MONDO)').alias('diseaseFromSourceId'),
+            array(col('MOI')).alias('allelicRequirements'),
+            array(col('urls')).alias('urls'),
+            col('CLASSIFICATION').alias('confidence'),
+            col('GCEP').alias('studyId')
+        )
+    )
+
+def write_evidence_strings(evidence: DataFrame, output_file: str) -> None:
+    '''Exports the table to a compressed JSON file containing the evidence strings'''
```
tskir, line 116:
```suggestion
    """Exports the table to a compressed JSON file containing the evidence strings"""
```

modules/ClinGen.py
```diff
@@ -1,137 +1,163 @@
...
     # Parse CLI arguments
     parser = argparse.ArgumentParser(description='Parse ClinGen gene-disease associations from Gene Validity Curations')
-    parser.add_argument('-i', '--input_file',
+    parser.add_argument('--input_file',
                         help='Name of csv file downloaded from https://search.clinicalgenome.org/kb/gene-validity',
                         type=str, required=True)
-    parser.add_argument('-o', '--output_file',
-                        help='Name of evidence file',
+    parser.add_argument('--output_file',
+                        help='Absolute path of the gzipped, JSON evidence file.',
```
tskir, line 135:
```suggestion
                        help='Absolute path of the gzipped JSON evidence file.',
```

tskir (dismissed):
Everything looks great! Thanks for adding uniform & proper PySpark initialisation to every parser.

I only have three very minor style suggestions (per PEP8 docstrings must always be in double quotes, not single), which I'll apply myself

tskir (approved):
Apparently I need to explicitly re-approve in order to make GitHub happy and not perform a forced merge, so doing exactly this

PR #106, opened 2021-11-12
Genetics debugging to pass validation + refactoring

ireneisdoomed:
Issues the parser was presenting and are now fixed:
- Evidence strings with "infinity" in the OR fields (see [#1617](https://github.com/opentargets/platform/issues/1617)). A filter is applied in OR values to set it to null when it cannot be represented with a single precision float. This is by using `sys.float_info.max`. I expect this to fix [#1687](https://github.com/opentargets/platform/issues/1687).
- Evidence strings without `diseaseFromSource`. Due to the presence of studies in the L2G table that were not in the study index. Fixed by changing the joining method to a `inner`between l2g and studies.
- Evidence strings without `diseaseFromSourceMappedId`. Due to the loss of records when the df was exploded on the trait_efos array column. When the array is empty, these studies are lost. I changed it to an explode_outer.
- One evidence string with an HANCESTRO ID in `diseaseFromSourceMappedId`. These records are filtered out.

There are many structural changes to the code in an effort of modularising all the logic that was handled in `main`, so that maintenance and future improvement of the code take less time to develop. In future PRs, my intention is to move the initialisation of the logger + spark session to a common module.

**To align with the rest of the parsers, the output of this script is a single JSON file (not a directory).**


DSuveges (changes requested):
Let's discuss the points I raised, might or might not be an issue.

PR #107, opened 2021-12-14
Update gene2phenotype parser for new field names.

DSuveges:
The release of Gene2Phenotypes dataset is out. There are a few changes that needed to address:
* Columns got renamed, new columns got introduced. 
* Variant consequences got changes. As the script maps consequence terms to SO codes, the mapping function had to be updated.

The script at this point runs just fine, validates no problem against the most recent schema. Example of the output:

```json
{
  "diseaseFromSource": "Childhood-Onset, Slowly Progressive Nemaline Myopathy",
  "datasourceId": "gene2phenotype",
  "datatypeId": "genetic_literature",
  "targetFromSourceId": "MYPN",
  "confidence": "both RD and IF",
  "studyId": "DD",
  "literature": [
    "28017374"
  ],
  "allelicRequirements": [
    "biallelic_autosomal"
  ],
  "variantFunctionalConsequenceId": "SO_0002317"
}
```

### DIFF
--- a/modules/Gene2Phenotype.py
+++ b/modules/Gene2Phenotype.py
@@ -13,18 +13,14 @@ from pyspark.sql.types import IntegerType, StringType, StructType, TimestampType
 from common.ontology import add_efo_mapping
 from common.evidence import detect_spark_memory_limit, write_evidence_strings
 
+
 G2P_mutationCsq2functionalCsq = {
-    'loss of function': 'SO_0002054',  # loss_of_function_variant
-    'all missense/in frame': 'SO_0001650',  # inframe_variant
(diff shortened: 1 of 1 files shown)


ireneisdoomed (dismissed):
Updates are in accordance with [G2P's changelog](https://www.ebi.ac.uk/gene2phenotype/updates_to_our_terms). New variant consequences are also correctly mapped to SO.

The script outputs 4244 evidence strings, which account for 2308 t/d associations - the same amount of associations we currently have in production.

I'm pushing 2 minor commits to make the spark instance a global variable; and the flag to make the input files required by the argument parser.

ireneisdoomed (approved):

PR #108, opened 2021-12-20
Updating the TEP parser

DSuveges:
Thank to the communication initiated by @andrewhercules , the [Structural Genomics Consortium](https://www.thesgc.org/) made the Target Enabling Packages (TEPs) available through a downloadable tsv file. This allowed us the following changes:

* Instead of parsing fragile html files, we can directly read data from a properly formatted table.
* We could increase consistency across other datasource parsers by using PySpark in favour of Pandas. 
* Shared modules used imported from `common.evidence`

The stable url for the data is hardcoded into the script. I think adding to the configuration does not make sense at this point.

### DIFF
--- a/modules/TEP.py
+++ b/modules/TEP.py
@@ -5,12 +5,14 @@ import argparse
 import logging
 import sys
 
-from common.evidence import detect_spark_memory_limit, write_evidence_strings
 from pyspark import SparkFiles
 from pyspark.conf import SparkConf
 from pyspark.sql import SparkSession
 from pyspark.sql.functions import (col, concat, explode, lit, lower,
                                    regexp_replace, split)
+
+from common.evidence import detect_spark_memory_limit, write_evidence_strings
+
 
 # The TEP dataset is made available by the SGC as a tab-separated file:
 TEPURL = 'https://www.thesgc.org/sites/default/files/fileuploads/available-teps.tsv'
--- a/modules/TEP.py
+++ b/modules/TEP.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-"""Parser for Target Enabling Pacages (TEP)."""
+"""Parser for Target Enabling Packages (TEP) downloaded from the Structural Genomics Consortium website."""
 
 import argparse
 import logging
@@ -13,8 +13,6 @@ from pyspark import SparkFiles
 from common.evidence import detect_spark_memory_limit, write_evidence_strings
 
 
-'''
-This script retrieves TEP (target enabling package) from the structural genomics consoritum website.
 '''
 
 # The TEP dataset is made available by the SGC as a tab-separated file:
@@ -45,7 +43,7 @@ def main(outputFile: str) -> None:
 
     # Fetching and processing the TEP table and saved as a JSON file:
     TEP_df = (
-        spark.read.csv(SparkFiles.get("available-teps.tsv"), sep='\t', header=True)
+        spark.read.csv(SparkFiles.get(TEPURL.split('/')[-1]), sep='\t', header=True)
 
         # Generating TEP url from Gene column: SLC12A4/SLC12A6 -> https://www.thesgc.org/tep/SLC12A4SLC12A6
         .withColumn('url', concat(lit('https://www.thesgc.org/tep/'), regexp_replace(lower(col('Gene')), '/', '')))
--- a/modules/TEP.py
+++ b/modules/TEP.py
@@ -1,76 +1,70 @@
-import pandas as pd
-import requests
-from bs4 import BeautifulSoup, UnicodeDammit
+#!/usr/bin/env python3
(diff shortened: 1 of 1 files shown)


modules/TEP.py
```diff
@@ -1,76 +1,74 @@
...
-        logging.error(f'The following symbols were not unique: {duplicated_targets}')
-        raise ValueError('The target symbol list is expected to be unique! Check logs for deatils.')
+    logging.info('TEP dataset has been downloaded and formatted.')
+    logging.info(f'Number of TEPs: {TEP_df.count()}')
+    logging.info(f'Number of unique genes: {TEP_df.select("targetFromSourceId").distinct().count()}')
 
     # Saving data:
-    logging.info(f'Saving data to {outputFile}.')
-    tep_list.to_json(outputFile, compression='infer', orient='records', lines=True)
+    write_evidence_strings(TEP_df, outputFile)
```
ireneisdoomed, line 70: Since we are no longer talking of evidence strings, I'd suggest to change the name of this function to something like `write_json_strings`/`write_df_to_json`/`export_to_json`. I'd also change the name of `common.evidence.py` to `common.utils.py`
```suggestion
    write_json_strings(TEP_df, outputFile)
```

modules/TEP.py
```diff
@@ -1,76 +1,74 @@
...
+        .set('spark.driver.memory', f'{spark_mem_limit}g')
+        .set('spark.executor.memory', f'{spark_mem_limit}g')
+        .set('spark.driver.maxResultSize', '0')
+        .set('spark.debug.maxToStringFields', '2000')
+        .set('spark.sql.execution.arrow.maxRecordsPerBatch', '500000')
+    )
+    spark = (
+        SparkSession.builder
+        .config(conf=spark_conf)
+        .config("spark.sql.broadcastTimeout", "36000")
```
ireneisdoomed, line 39: What is the function of this setting? Data is not broadcasted as far as I can tell. Script worked when I turned it off.

modules/TEP.py
```diff
@@ -1,76 +1,74 @@
-import pandas as pd
-import requests
-from bs4 import BeautifulSoup, UnicodeDammit
+#!/usr/bin/env python3
+"""Parser for Target Enabling Pacages (TEP)."""
```
ireneisdoomed, line 2:
```suggestion
"""Parser for Target Enabling Packages (TEP) downloaded from the Structural Genomics Consortium website."""
```

modules/TEP.py
```diff
@@ -1,76 +1,74 @@
...
+from common.evidence import detect_spark_memory_limit, write_evidence_strings
 
-        return tep_url
 
-    def get_therapeutic_area(row):
-        therapeutic_area = row.findAll('td')[2].text
-        return therapeutic_area
-
-    URL = 'https://www.thesgc.org/tep'
+'''
```
ireneisdoomed, line 16:
```suggestion
```

modules/TEP.py
```diff
@@ -1,76 +1,74 @@
...
+This script retrieves TEP (target enabling package) from the structural genomics consoritum website.
```
ireneisdoomed, line 17:
```suggestion
```

modules/TEP.py
```diff
@@ -1,76 +1,74 @@
...
+'''
```
ireneisdoomed, line 18:
```suggestion
```

ireneisdoomed (changes requested):
Parser works great! There are no discrepancies with the previos scraped version and the output validates against the schema. 

I've simply made some minor comments. Besides that, it would be nice to adapt the README so that it is not so evidence-focused. We should also start to think of a better name for the repo 🙄 `platform_data_parsers`?

modules/TEP.py
DSuveges: I have practically lifted over the spark initialisation steps from other scripts. It's a relevant question though...

modules/TEP.py
DSuveges: I 100% with this call, however let's merge this version as it is, and have an other PR with a ticket so this update would be more transparent.

modules/TEP.py
ireneisdoomed: Apparently this configuration is only in this parser. Since it works perfectly without it, I suggest to change it and set this setting when we unify the spark instantiation in common.utils.

modules/TEP.py
```diff
@@ -5,14 +5,12 @@
 import logging
 import sys
 
+from common.evidence import detect_spark_memory_limit, write_evidence_strings
```
ireneisdoomed, line 8: Local modules come last when sorting imports, at least according to PEP. Are you using any type of formatter?

DSuveges: Yes, that's right. I'll fix it.

ireneisdoomed (approved):

PR #109, opened 2022-01-11
Add safetyLiabilities module

ireneisdoomed:
This PR closes [#1872](https://github.com/opentargets/platform/issues/1872).

This module contains a particularity that I've been reading about during Christmas. Apparently `argparse` is old school and other options are being used consistently (Click, Typer and Python Fire, being the last 2 more dominant).

I've been playing with Typer and Fire. The main reason to start using any of this is pure simplicity. You will see in the code how incredibly simple it is to set up a quick argument parser (literally one line). So it is easy to use it at least for prototyping.

[Typer](https://typer.tiangolo.com) is very potent, their main point being that it uses python's standard type hints as validation for user input.
I went for Fire because the implementation was slightly simpler and because of a very interesting feature, the possibility of entering an interactive mode (see [docs](https://github.com/google/python-fire/blob/master/docs/guide.md))


ireneisdoomed:
Changes:
- [X] New module has been added to Snakemake.
- [X] Added validation step.
- [X] Evidence is no longer grouped on the target. This step will still be part of the ETL, so that grouping will happen once all targets are resolved. There have not been any changes to the schema, so that the implementation for BE will be very easy.
- [X] I've gone back to using `argparse`. I've been thinking after our conversation this morning and I don't really have a grounded reason to add the dependency other than it being clean. I'll keep using it for prototyping.

All the JSON strings validate against the new schema. This is an example:
```
{
  "id": "ENSG00000168398",
  "event": "hypertension",
  "datasource": "Urban et al. (2012)",
  "eventId": "EFO_0000537",
  "url": "https://doi.org/10.1002/9781118098141.ch2",
  "biosample": {
    "tissueLabel": "cardiovascular system",
    "tissueId": "UBERON_0004535"
  },
  "effects": [
    {
      "direction": "inhibition",
      "dosing": "general"
    }
  ]
}
```

**Important**. The correct functioning of this code is dependant of 2 PRs:
- The one that organises the target safety curation ([#2](https://github.com/opentargets/curation/pull/2))
- The one for the Safety JSON schema ([#138](https://github.com/opentargets/json_schema/pull/138))

modules/TargetSafety.py
```diff
@@ -0,0 +1,305 @@
...
+        '--toxcast',
+        help='Input table containing biological processes associated with relevant targets that have been observed in toxicity assays.',
+        type=str,
+        required=True,
+    )
+    parser.add_argument(
+        '--adverse_events',
+        help='Input TSV containing adverse events associated with targets that have been collected from relevant publications. Fetched from GitHub..',
+        type=str,
+        required=True,
```
DSuveges, line 279: I think this is not good. If the github file is hardcoded, this argumentum isn't required. However, I would rather keep this argument mandatory and remove the hardcoded variable and add the path to the config file.

modules/TargetSafety.py
```diff
@@ -0,0 +1,305 @@
...
+    )
+    parser.add_argument(
+        '--safety_risk', help='Input TSV containing cardiovascular safety liabilities associated with targets that have been collected from relevant publications. Fetched from GitHub.', type=str, required=True
```
DSuveges, line 282: Same as the above argument.

modules/TargetSafety.py
```diff
@@ -0,0 +1,305 @@
...
+    return parser
+
+if __name__ == '__main__':
+    args = get_parser().parse_args()
+    
+    main(
+        toxcast=args.toxcast,
+        adverse_events=args.adverse_events,
+        safety_risk=args.safety_risk,
+        output=args.output
```
DSuveges, line 304: If the arguments are passed like this, the `safety_risk` will always be submitted even if the script was not called with safety risk file as being optional. This have a side effect: the default value defined in row 28,29. But if the other comments are sorted, this will not be an issue.

DSuveges (changes requested):
When circumventing the issues I highlighted in my comments, the script runs and generates 5775 safety object, which successfully validates against the most recent schema.

Example:
```json
{
  "id": "ENSG00000168398",
  "event": "arteriolar dilation/hypotension",
  "datasource": "Urban et al. (2012)",
  "eventId": "EFO_0005251",
  "url": "https://doi.org/10.1002/9781118098141.ch2",
  "biosample": {
    "tissueId": "UBERON_0004535",
    "tissueLabel": "cardiovascular system"
  },
  "effects": [
    {
      "direction": "activation",
      "dosing": "general"
    }
  ]
}
```

DSuveges (approved):

PR #111, opened 2022-01-25
New module: ChEMBL.py

ireneisdoomed:
This module:
1. Reads a ChEMBL submission
2. Reads the reasons to stop categorisation outputted by running https://github.com/ireneisdoomed/stopReasons
3. Merge both data and exports a new set of evidence that follow the proposed JSON schema https://github.com/opentargets/json_schema/pull/140

Snakemake rules are working. Output can be found in a new location: `gs://otar000-evidence_input/ChEMBL/json/chembl-2022-01-25.json.gz`
For now I have hardcoded the branch of the schema to use, everything validates fine.

One question: arguments of the parser are rather dynamic. We'll have to pick the latest ChEMBL evidence and the latest corresponding predictions. When specifying these locations in the config file, neither regex or the * wildcard in the filenames worked. I also read about Snakemake's [wildcards](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#wildcards) but those didn't work as well. Do you know how these can be specified in a way that we don't have to change the config every time we run it?

~Same as the Schema PR, I am drafting this since I am still working on the notebooks to figure out the scoring.~


DSuveges (dismissed):
OK, parser runs, the new fields are successfully added to the evidence data sourced from chembl. All looks good.

DSuveges (approved):

PR #120, opened 2022-02-16
Updating the logging design of the evidence generation pipeline

tskir:
:warning: Built on top of #119 and includes all changes from it as well.

Closes https://github.com/opentargets/platform/issues/1914.

Log generation is now local and static for each individual parser. Everything's uploaded and timestamped in the `all` step.


tskir:
Both local & upload tests have now been completed. Everything has been tested and ready for reviews

Snakefile
```diff
@@ -139,26 +161,25 @@ rule crispr:
...
-        GS.remote(logFile)
+        'log/epmc.log'
     shell:
         """
+        exec &> {log}
         python modules/EPMC.py \
           --cooccurrenceFile {input.inputCooccurences} \
           --output {output.evidenceFile}
         opentargets_validator --schema {params.schema} {output.evidenceFile}
         """
```
DSuveges, line 180: So the pipeline is still failing. The problem is still the fact that the input is a directory which is not supported for inputs. As a quick workaround, I would suggest that we specify the input as a parameter. 

```python
rule epmc:                    # Process target/disease evidence strings from ePMC cooccurrence files.
    params:
        inputCooccurences = directory(GS.remote(config['EPMC']['inputBucket'])),
        schema = f"{config['global']['schema']}/opentargets.json"
    output:
        evidenceFile = 'epmc.json.gz'
    log:
        'log/epmc.log'
    shell:
        """
        exec &> {log}
        python modules/EPMC.py \
          --cooccurrenceFile {input.inputCooccurences} \
          --output {output.evidenceFile}
        opentargets_validator --schema {params.schema} {params.evidenceFile}
        """
```

As parameters doesn't go through such a checks.

Snakefile
```diff
@@ -168,16 +189,17 @@ rule gene2Phenotype:
...
+        evidenceFile = 'gene2phenotype.json.gz'
     log:
-        GS.remote(logFile)
+        'log/gene2Phenotype.log'
     shell:
         """
-        # Retain the inputs and save to GCS.
+        exec &> {log}
+        # Retain the inputs. They will be later uploaded to GCS.
         cp {input.ddPanel} {output.ddBucket}
```
DSuveges, line 203: Unfortunately as `GS.remote` was removed from the input, `cp` command doesn't work anymore.

We have to use either `curl` or `wget` eg:
```bash
wget ftp.ebi.ac.uk/pub/databases/gene2phenotype/DD/DDG2P_1_12_2021.csv.gz -O DDG2P.csv.gz
```

DSuveges: Also, when calling `modules/Gene2Phenotype.py` the input files needs to be changed:

```bash
        python modules/Gene2Phenotype.py \
          --dd_panel {output.ddBucket} \
          --eye_panel {output.eyeBucket} \
          --skin_panel {output.skinBucket} \
          --cancer_panel {output.cancerBucket} \
          --output_file {output.evidenceFile} \
          --cache_dir {params.cacheDir} \
          --local
```

Snakefile
tskir: This turned out to be a bug in Snakemake, resolved with updating to v7.3.0. In practice, even though the `directory` flag doesn't affect the input section (and in that sense, is "not allowed" there), the actual directory download is still being processed correctly

Snakefile
tskir: Actually I wrote a reply to this one a week ago, at the same time as the previous comment, but apparently I forgot to send it :man_facepalming: 

The problems mentioned here were also caused by the same Snakemake bug when handling the remote files. Everything works now: the rule **gene2Phenotype** copies the remove files (DD, eye, skin, cancer) to local ones; then the rule **all** uploads them to retain in Google Storage, alongside all other files listed in `ALL_FILES`

DSuveges (approved):
All looks and work fine.

PR #122, opened 2022-04-13
22.04 configuration

ireneisdoomed:
This PR contains:
- renaming of Phenodigm to IMPC ([#2551](https://github.com/opentargets/issues/issues/2551))
- addition of `variantFunctionalConsequenceId` To Orphanet ([#2555](https://github.com/opentargets/issues/issues/2555))
- bug fix in the function that adds the EFO mapping ([#2556](https://github.com/opentargets/issues/issues/2556))
- updates in the config for 22.04 data generation
- Small bug fix in the CRISPR parser
- Added PMID:30104761 as a reference for SAIGE studies in Genetics Portal
- Added `UK Biobank 500k` as `cohortId` for SAIGE and NEALE studies in Genetics Portal. This makes the data more analogous to `gene_burden`
- Updated the probes generation notebook. This is still WIP ([#2557](https://github.com/opentargets/issues/issues/2557))

PR #121, opened 2022-04-07
Parse AZ PheWAS Portal and Regeneron data for `gene_burden`

ireneisdoomed:
This PR contains:
- The module that processes data from REGENERON: `RegeneronGeneBurden.py`. Input files are their raw data (`gs://otar000-evidence_input/GeneBurden/data_files/41586_2021_4103_MOESM5_ESM.xlsx`) and the GWAS Catalog studies. The GWAS Catalog file is uploaded to the `data_files` bucket by default so that we can version each run.
- The module that processes data from the AZ PheWAS Portal: `AzGeneBurden.py`. Input files are 2 parquet directories with the binary and quantitative traits (also found at `gs://otar000-evidence_input/GeneBurden/data_files`)
- The module that calls the previous two and brings their output together: `GeneBurden.py`. Input files are the above mentioned.
- New `GeneBurden` rule in the pipeline and its configuration.

The latest version can be found at `gs://otar000-evidence_input/GeneBurden/json/gene_burden-2022-04-07.json.gz`

At the moment I think practically everything is ready. The validator, though, is complaining because of some targets not following the pattern:

1. Most of the offending ones are cases like `RP11-545J16.1`. I think this is a synonym for [ENSG00000257046](https://platform.opentargets.org/target/ENSG00000257046/associations), but I think we wouldn't pick it up in the ETL.
2. We have examples of legit gene names that should be passing validation `HLA-DRB5` and that are in the index. In this case the problem is that the enforced pattern doesn't allow for strings with special characters. I guess we should change this.
3. The jewel in the crown is `2021-03-08 00:00:00`. REGENERON data comes in an Excel file, so we have records with the known issue of interpreting the gene name as a date.

Closes [#2545](https://github.com/opentargets/issues/issues/2545)

PR #125, opened 2022-05-06
Il 22.04

ireneisdoomed:
This PR includes minor changes in the `22.04` branch since #122.


modules/EPMC.py
```diff
@@ -4,160 +4,171 @@
...
-            (pf.col('publicationIdentifier').isNotNull()) &  # Making sure at least the pmid or the pmcid is given:
-            (pf.length(pf.col('text')) < 600) &  # Exclude sentences with more than 600 characters
-            (pf.col('label1').isin(EXCLUDED_TARGET_TERMS) == False)  # Excluding target labels from the exclusion list
+            (F.col('type') == 'GP-DS')
+            & (F.col('isMapped'))  # Filter gene/protein - disease cooccurrence
+            & (F.col('publicationIdentifier').isNotNull())  # Filtering for mapped cooccurrences
+            & (F.length(F.col('text')) < 600)  # Making sure at least the pmid or the pmcid is given:
+            & (  # Exclude sentences with more than 600 characters
+                F.col('label1').isin(EXCLUDED_TARGET_TERMS) == False
+            )  # Excluding target labels from the exclusion list
```
DSuveges, line 83: Comments got shifted. Also I have a feeling the brackets are excessively used here:
```suggestion
            (F.col('type') == 'GP-DS')  # Filter gene/protein - disease cooccurrence
            & F.col('isMapped')  # Filtering for mapped cooccurrences
            & F.col('publicationIdentifier').isNotNull(). # Making sure at least the pmid or the pmcid is given:
            & (F.length(F.col('text')) < 600) # Exclude sentences with more than 600 characters
            & (F.col('label1').isin(EXCLUDED_TARGET_TERMS) == False)  # Excluding target labels from the exclusion list
```

modules/ChEMBL.py
```diff
@@ -30,7 +30,7 @@ def main(chembl_evidence: str, predictions: str, output_file: str) -> None:
     logging.info(f'Classes of reason to stop table: {predictions}')
 
     # Load input into dataframes
-    chembl_df = spark.read.json(chembl_evidence).persist()
+    chembl_df = spark.read.json(chembl_evidence).repartition(20).persist()
```
DSuveges, line 33: Why do you need repartitioning the data?

configuration.yaml
```diff
@@ -8,12 +8,7 @@ global:
 cancerBiomarkers:
   inputBucket: gs://otar000-evidence_input/CancerBiomarkers/data_files
   outputBucket: gs://otar000-evidence_input/CancerBiomarkers/json
-  # Downloaded separately using wget, because FTP.RemoteProvider cannot handle recursive directory downloads.
-  drugIndex: ftp.ebi.ac.uk/pub/databases/opentargets/platform/22.02/output/etl/parquet/diseases
-ChEMBL:
-  evidence: gs://otar008-chembl/cttv008-20-01-2022.json.gz
-  stopReasonCategories: gs://otar000-evidence_input/ChEMBL/data_files/chembl_predictions-2022-01-24.tsv
-  outputBucket: gs://otar000-evidence_input/ChEMBL/json
+  drugIndex: gs://open-targets-pre-data-releases/22.04/output/etl/parquet/molecule
```
DSuveges, line 11: I see there's an other section of ChEMBL. But why are you deleting this one and keeping the one, which is out of the alphabetic order?

common/evidence.py
```diff
@@ -50,3 +50,42 @@ def initialize_sparksession() -> SparkSession:
     )
 
     return spark
+
+
+def read_path(path: str, spark_instance) -> DataFrame:
```
DSuveges, line 55: Are we sure we need such an elaborate mechanism to to handle paths? Also what this function does is not exactly "reading path" in my opinion.

DSuveges (dismissed):
Mostly agree with the proposed modifications, just some comments here and there. Please take a look. Nothing dealbreaker. The only thing I'm not exactly convinced about is this `read_path` function.

DSuveges:
Why don't we have travis set up for this repo? I'm sure there used to be, but it has been removed...

modules/ChEMBL.py
ireneisdoomed: When reading data from JSON, Spark loads everything in the main node rather than splitting it across multiple workers. Since ChEMBL evidence is >300k records, I thought it would be beneficial to redistribute that chunk of data in different nodes to make the join more performant.

That was the theory. Reality is this:
```python3
In [34]: timeit -n 3 spark.read.json(chembl).join(predictions_df, on='studyStopReason', how='left').distinct()
2.48 s ± 8.3 ms per loop (mean ± std. dev. of 7 runs, 3 loops each)

In [35]: timeit -n 3 spark.read.json(chembl).repartition(20).join(predictions_df, on='studyStopReason', how='left').distinct()
2.48 s ± 12.1 ms per loop (mean ± std. dev. of 7 runs, 3 loops each)
```

No actual benefit when repartitioning. Thanks for flagging, Ill remove the statement.

common/evidence.py
ireneisdoomed: It's a one stop shop function that we use in the metrics repo to be able to read any sort of input.
I thought we could benefit from having such a common function that we could make use of in every parser. Please let me know your concerns.
About the name, would `read_path_to_df` be better?

ireneisdoomed:
> Why don't we have travis set up for this repo? I'm sure there used to be, but it has been removed... https://github.com/opentargets/evidence_datasource_parsers/pull/121 still had, but it has been removed. Do you know by any chance what the justification was?

As discussed in the SU, the justification was that the Travis set up was pointing to a dummy directory where we are supposed to have tests. As we agreed that having that dummy test script didn't make sense, I also decided to remove the Travis conf.
However, we have agreed to be alert of any tests we could implement in future iterations so that we could make good use of Travis.

configuration.yaml
ireneisdoomed: Fixed in https://github.com/opentargets/evidence_datasource_parsers/pull/125/commits/689142fa8dc1ebd6be9d24b9b41bdb4e5ea5ed4f.

common/evidence.py
DSuveges: Nah, it's not really important. In a way it makes sense. We can keep using it as it is.

DSuveges (approved):
All looks good. thank you for addressing my questions.

PR #126, opened 2022-06-01
Updated pathways to Reactome mappings

ireneisdoomed:
This PR updates the pending mappings between the perturbed pathways and their Reactome IDs that are exposed in the PROGENy evidence as described in [#713](https://github.com/opentargets/issues/issues/713)

Evidence looks fine and passes validation.


DSuveges (approved):
All looks good.

PR #132, opened 2022-08-08
Il-2687

ireneisdoomed:
This PR includex:
- ~An assertion to check that bugs like the one reported in #2687 won't happen again~
- An improvement in the way the spark session is handled. It used to be an argument of many functions and now we simply call the method that gets the session that is current at runtime
- A clearer way of defining Pyspark functions and types


modules/GeneBurden.py
```diff
@@ -45,74 +41,94 @@ def main(
...
+                (F.col('projectId') == 'Autism Sequencing Consortium')
+                & (F.col('diseaseFromSourceMappedId') != 'EFO_0003756')
+            )
+            | ((F.col('projectId') == 'SCHEMA consortium') & (F.col('diseaseFromSourceMappedId') != 'MONDO_0005090'))
+            | ((F.col('projectId') == 'Epi25 collaborative') & (F.col('diseaseFromSourceMappedId') != 'MONDO_0100062'))
+        ).count()
+        != 0
+    ):
+        logging.exception('Curated evidence is not mapped to the correct disease.')
+        raise AssertionError('Curated evidence is not mapped to the correct disease.')
```
DSuveges, line 90: I have a bad feeling about this... in a way you are checking the manual disease mappings at evidence level, and I'm not sure if this is good. What if EFO changes the autism spectrum disorder to MONDO_0014248 (also autism spectrum disorder) this assertion would pass, but because the mapping is outdated, the evidence would pointing to a wrong disease.

modules/GenebassGeneBurden.py
```diff
@@ -50,13 +51,13 @@ def main(genebass_data: str, spark_instance: SparkSession) -> DataFrame:
     # Write output
     evd_df = parse_genebass_evidence(genebass_df)
 
-    if evd_df.filter(col('resourceScore') == 0).count() != 0:
-        logging.error('There are evidence with a P value of 0.')
+    if evd_df.filter(F.col('resourceScore') == 0).count() != 0:
+        logging.assertion('There are evidence with a P value of 0.')
```
DSuveges, line 55: I'm not sure if `logging.assertion` is a thing.. have you tried this?

DSuveges (changes requested):
Mostly things makes sense. Changing how we are handling spark sessions across functions and by importing the import of spark functions and classes made things more sleek and nicer. However I'm not sure about the assertion... for me, I hardly see the point, also I couldn't reproduce the functionality of `logging.assertion`. Could you please taking a look, and we can discuss the potential ways of validating the disease mappings.

modules/GenebassGeneBurden.py
ireneisdoomed: Thank you for flagging this. I got it confused with `logging.exception`

modules/GeneBurden.py
ireneisdoomed: We have agreed that the evidence generation is not the right place to do this checks. Thank you so much for the input ^^

DSuveges (approved):
All looks good. These updates were discussed earlier.

PR #133, opened 2022-08-11
Add year of publication to the EPMC evidence

ireneisdoomed:
Each EPMC evidence describes the t/d relationship in a given publication. We want to make the users able to filter the evidence by the publication year.

This PR processes the `year` field available in the cooccurrences to expose that in `publicationYear`. 
There are cases where more than one date is reported for the same publication. In this case we take the earlier year as the publication date (I checked some examples and this was accurate).


DSuveges (approved):
All good. However if we could store publication date, that would be even better.

PR #136, opened 2022-08-30
Updating pyspark version.

DSuveges:
Dependabot found a problematic pyspark version and recommended to change pyspark version to `3.1.3`. I didn't merge [dependabot's  PR](https://github.com/opentargets/evidence_datasource_parsers/pull/135), instead updated the requirements and requirements-frozen properly. 

Before merging, we have to make sure no pyspark functionality is broken. (the full pipeline needs to be re-run anyways.)

### DIFF
--- a/requirements-frozen.txt
+++ b/requirements-frozen.txt
@@ -29,6 +29,7 @@ google-resumable-media==2.3.3
 googleapis-common-protos==1.56.4
 httplib2==0.20.4
 idna==3.3
+importlib-resources==5.9.0
 Jinja2==3.1.2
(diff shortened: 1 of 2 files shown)


tskir:
I've re-ran the entire pipeline and checked for changes in the output files.

The pipeline ran without errors.

Nearly all files are entirely unchanged (same md5sum after uncompressing and sorting).

The only two files which were changes had changes which were expected:
* `genomics_england.json.gz` has added `literature` fields
* `clingen.json.gz` contains one single additional record (probably added upstream)

tskir (approved):

PR #139, opened 2022-09-23
Remove EPMC evidence generation

ireneisdoomed:
This PR the steps to generate EPMC's evidence from our pipeline. This logic has been moved to the ETL.


DSuveges (approved):
I'm sad to see some of my first pyspark script go... but it's for the better. We knew the annoying circular dependency from the first second, but at that time we had to make this compromise.

PR #141, opened 2022-11-07
Remove duplicated evidence from IMPC

tskir:
Closes https://github.com/opentargets/issues/issues/1805.

The following fields are now used to enforce uniqueness:
```python
    # Specific to IMPC.
    'diseaseFromSource',  # Original disease name.
    'targetInModel',  # Mouse gene name.
    'biologicalModelAllelicComposition',  # Mouse model property.
    'biologicalModelGeneticBackground',  # Mouse model property.

    # General.
    'diseaseFromSourceMappedId',  # EFO mapped disease ID.
    'targetFromSourceId',  # Ensembl mapped human gene ID.
```

Evidence count decrease is now adequate, in its expected order of magnitude, to our expectations.

### DIFF
--- a/modules/IMPC.py
+++ b/modules/IMPC.py
@@ -38,6 +38,12 @@ IMPC_SOLR_TABLES = {
 
 # List of fields on which to enforce uniqueness by only keeping the record with the highest score.
 UNIQUE_FIELDS = [
+    # Global
+    'targetId',
+    'targetFromSourceId',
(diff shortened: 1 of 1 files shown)


tskir:
The code and data are located on the instance `tskir-evidence-generation` if anyone would like to check it out

DSuveges (approved):
Everything makes perfect sense.

PR #153, opened 2023-01-27
Minor fixes on the Safety and Burden parsers

ireneisdoomed:
- Aopwiki data needed an extra parsing step to leave it in the same shape as the rest of the sources before grouping;
- Burden curation table had one column changed its order, making the schema in the parser outdated.


DSuveges (approved):
All looks good.

PR #155, opened 2023-02-02
Update dependency versions to resolve security issues

tskir:
Closes https://github.com/opentargets/issues/issues/2839. 

## Quick summary

The dependencies are now updated and the result thoroughly tested. The process was complicated and delayed by several other issues which were discovered (and some of them also fixed) while working on it.

Should similar work be required in the future, this will be a much quicker process because I won't have to go through all the same problems.

### Untangling transitive dependencies (three levels deep...)

In total, there were three vulnerable dependencies. Two of them, `GitPython` and `future`, could be simply updated to newer versions. However, the third one, `py`, is a deprecated project that will never receive further updates.

While we didn't rely on `py`, we relied on the `retry` module, which in turn relied on `py`. However, `retry` could not be updated because it is _also_ in inactive development state and will never receive further updates.

To resolve this, I replaced `retry` with its actively maintained fork `retry2`, which reimplements the library and does not depend on `py` anymore.

This didn't help though, because this repository also depends on `ontoma`, which depends on `retry`, which depends on `py`. To mitigate this, I had to update OnToma dependencies as well and release a new version [v1.1.0](https://github.com/opentargets/OnToma/releases/tag/v1.1.0).

### Testing the updates

Because a lot of dependencies were updated, I had to make sure that this doesn't cause any unintended consequences. This is a serious matter. For example, a few years ago Pandas shipped a very unstable release with more than 20 critical breaking bugs, which would affect the output severely.

Testing the updates didn't prove easy, however. First, it turned out that the `opentargets-validator` sometimes hangs indefinitely and doesn't allow the pipeline to complete. In this particular instance it was circumvented by fixing the offending target safety evidence strings, but the issue will need to be addressed, for which I created a ticket https://github.com/opentargets/issues/issues/2855.

Additionally, file comparisons were complicated by the fact that not all parsers ensure stable sort order of their outputs, for which I also created a (lower priority) ticket: https://github.com/opentargets/issues/issues/2858.

### Repository cleanup

While I was at it, I went through the repository and removed all obsolete and unused files and modules, and updated `.gitignore` rules.


DSuveges (approved):
All looks sensible. Thank you for taking a look for files we no longer use.

PR #157, opened 2023-04-11
Ds 2911 ppp snakemake

DSuveges:
Although the snakemake pipeline might not last too long, it makes sense to write up the rules for generating evidence files for PPP. 
- If the snakemake was at some point replaced by other methods, if the rules are available it would easier to make the transition.
- So far there was no documentation on this process. 

**Caveats**:

- The encore datasets are not yet moved to google storage.
- The management of the configuration is not yet fully solved: files are stored in a versioned, private repo, but hard to refer 
and access in a pipeline.
- The standard service account does not yet have write permission on the `gs://OTAR013-PPP/` bucket.  

Issue: [#2911](https://github.com/opentargets/issues/issues/2911)

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -43,6 +43,13 @@ ALL_FILES = [
 LOCAL_FILENAMES = [f[0] for f in ALL_FILES]
 REMOTE_FILENAMES = [f[1] for f in ALL_FILES]
 
+# At this point parsing the PPP datasources are not yet as integrated as to add these files to the complete set of output:
+PPP_sources = [
+    ('ot_crispr.json.gz', GS.remote(f"{config['OT_CRISPR']['outputBucket']}/ot_crispr-{timeStamp}.json.gz")),
+    ('validation_lab.json.gz', GS.remote(f"{config['ValidationLab']['outputBucket']}/validation_lab-{timeStamp}.json.gz")),
+    ('encore.json.gz', GS.remote(f"{config['Encore']['outputBucket']}/encore-{timeStamp}.json.gz"))
+]
+
 # Auxiliary rules.
 rule help:                    # Print help comments for Snakefile.
     input: "Snakefile"
@@ -427,7 +434,29 @@ rule ot_crispr:               # Generating evidence for OTAR CRISPR screens
             --study_table {params.study_table} \
             --data_folder {params.data_folder} \
             --output {output}
-        opentargets_validator --schema {params.schema} {output}
+        # opentargets_validator --schema {params.schema} {output}
+        """
+# --- Evidence generation for PPP --- #
+rule encore:               # Generating evidence for OTAR CRISPR screens
+    params:
+        data_folder = config['Encore']['data_directory'],
+        config = config['Encore']['config'],
+        schema = f"{config['global']['schema']}/opentargets.json"
+    output:
+        'encore.json.gz'
+    input:
+        cell_passport_table = HTTP.remote(config['global']['cell_passport_file'], keep_local=True),
+    log:
+        'log/encore.log'
+    shell:
+        """
+        exec &> {log}
+        python partner_preview_scripts/encore_parser.py \
+            --output_file {output} \
+            --parameter_file {params.config} \
+            --data_folder {params.data_folder} \
+            --cell_passport_file {input.cell_passport_table}
+        # opentargets_validator --schema {params.schema} {output}
         """
 
 rule validation_lab:               # Generating evidence for OTAR CRISPR screens
@@ -436,7 +465,7 @@ rule validation_lab:               # Generating evidence for OTAR CRISPR screens
         config = config['ValidationLab']['config'],
         schema = f"{config['global']['schema']}/opentargets.json"
     input:
-        cell_passport_table = HTTP.remote(config['ValidationLab']['cell_passport_file'], keep_local=True),
+        cell_passport_table = HTTP.remote(config['global']['cell_passport_file'], keep_local=True),
     output:
         'validation_lab.json.gz'
     log:
@@ -451,4 +480,13 @@ rule validation_lab:               # Generating evidence for OTAR CRISPR screens
             --output_file {output}
         opentargets_validator --schema {params.schema} {output}
         """
+# --- Moving local PPP files to remote --- #
+rule PPP:                          # Moving PPP evidence to destination bucket.
+    input:
+        [source[0] for source in PPP_sources]
+    output:
+        [source[1] for source in PPP_sources]
+    run:
+        for source in PPP_sources:
+            os.rename(source[0], source[1])
 
--- a/Snakefile
+++ b/Snakefile
@@ -430,3 +430,25 @@ rule ot_crispr:               # Generating evidence for OTAR CRISPR screens
         opentargets_validator --schema {params.schema} {output}
         """
 
+rule validation_lab:               # Generating evidence for OTAR CRISPR screens
+    params:
+        data_folder = config['ValidationLab']['data_directory'],
+        config = config['ValidationLab']['config'],
+        schema = f"{config['global']['schema']}/opentargets.json"
+    input:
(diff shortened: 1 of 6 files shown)


Snakefile
```diff
@@ -410,3 +417,76 @@ rule targetSafety:            # Process data from different sources that describ
             --output {output}
         opentargets_validator --schema {params.schema} {output}
         """
+# --- Evidence generation for PPP --- #
+rule ot_crispr:               # Generating evidence for OTAR CRISPR screens
```
tskir, line 421:
```suggestion

rule ot_crispr:               # Generating PPP evidence for OTAR CRISPR screens
```

tskir: (This is just to annotate rules in the same way that the rest are annotated)

Snakefile
```diff
@@ -410,3 +417,76 @@ rule targetSafety:            # Process data from different sources that describ
...
+    log:
+        'log/ot_crispr.log'
+    shell:
+        """
+        exec &> {log}
+        python partner_preview_scripts/ot_crispr.py \
+            --study_table {params.study_table} \
+            --data_folder {params.data_folder} \
+            --output {output}
+        # opentargets_validator --schema {params.schema} {output}
```
tskir, line 437: Is there a reason we don't want / cannot run the validation on PPP outputs?

Snakefile
```diff
@@ -410,3 +417,76 @@ rule targetSafety:            # Process data from different sources that describ
...
+        """
+# --- Evidence generation for PPP --- #
+rule encore:               # Generating evidence for OTAR CRISPR screens
```
tskir, line 440:
```suggestion

rule encore:               # Generating PPP evidence for ENCORE
```

Snakefile
```diff
@@ -410,3 +417,76 @@ rule targetSafety:            # Process data from different sources that describ
...
+        exec &> {log}
+        python partner_preview_scripts/encore_parser.py \
+            --output_file {output} \
+            --parameter_file {params.config} \
+            --data_folder {params.data_folder} \
+            --cell_passport_file {input.cell_passport_table}
+        # opentargets_validator --schema {params.schema} {output}
+        """
+
+rule validation_lab:               # Generating evidence for OTAR CRISPR screens
```
tskir, line 462:
```suggestion
rule validation_lab:               # Generating PPP evidence for Validation Lab
```

Snakefile
```diff
@@ -410,3 +417,76 @@ rule targetSafety:            # Process data from different sources that describ
...
+        exec &> {log}
+        python partner_preview_scripts/ValidationLab.py \
+            --parameter_file {params.config} \
+            --data_folder {params.data_folder} \
+            --cell_passport_file {input.cell_passport_table} \
+            --output_file {output}
+        opentargets_validator --schema {params.schema} {output}
+        """
+# --- Moving local PPP files to remote --- #
+rule PPP:                          # Moving PPP evidence to destination bucket.
```
tskir, line 484:
```suggestion

rule PPP:                          # Moving PPP evidence to destination bucket.
```

Snakefile
tskir, line 421:
```suggestion

rule ot_crispr:               # Generating PPP evidence for OTAR CRISPR screens
```

This is just to bring the style in alignment with existing rules, & leave a blank line

Snakefile
tskir, line 440:
```suggestion

rule encore:               # Generating PPP evidence for ENCORE
```

Snakefile
tskir, line 462:
```suggestion
rule validation_lab:               # Generating PPP evidence for Validation Lab
```

Snakefile
tskir, line 437: Is there a reason for why we don't want to / cannot run validation on PPP evidence?

Snakefile
tskir, line 484:
```suggestion

rule PPP:                          # Moving local PPP evidence to the destination bucket.
```

common/evidence.py
```diff
@@ -225,18 +242,46 @@ def read_path(path: str, spark_instance) -> DataFrame:
 def import_trait_mappings() -> DataFrame:
     """Load the remote trait mappings file to a Spark dataframe."""
 
-    remote_trait_mappings_url = (
-        'https://raw.githubusercontent.com/opentargets/curation/22.09.1/mappings/disease/manual_string.tsv'
-    )
+    remote_trait_mappings_url = "https://raw.githubusercontent.com/opentargets/curation/22.09.1/mappings/disease/manual_string.tsv"
```
tskir, line 245: I see that this PR only changes the style of this line, and not the contents, but I still have to ask: is there a reason this URL is hardcoded here, rather than being stored somewhere in a config?

configuration.yaml
```diff
@@ -72,3 +72,15 @@ TargetSafety:
   toxcast: gs://otar001-core/TargetSafety/data_files/toxcast/ToxCast_2021-08-17.tsv
   aopwiki: gs://otar001-core/TargetSafety/data_files/aopwiki/aopwiki-2023-01-20.json
   outputBucket: gs://otar001-core/TargetSafety/json
+OT_CRISPR:
+  data_directory: gs://otar013-ppp/ot_crispr/input_data
+  config: gs://otar013-ppp/PPP-evidencie-configuration/ot_crispr_config.json
+  outputBucket: gs://otar013-ppp/ot_crispr
+ValidationLab:
+  data_directory: /Users/dsuveges/project_data/validation_lab/2022.05/PPP_OTVL_May2022
```
tskir, line 80: I don't think this local path is supposed to be present here?

(Also the same for the Encore section)

tskir (changes requested):

Snakefile
DSuveges: Oh, no. Thanks for spotting! The validation is totally possible, however, i turned off for performance while prototyping.

common/evidence.py
DSuveges: No, and i think i have already raised to @ireneisdoomed that this practice needs to be reviewed. This path should come from the config.

configuration.yaml
DSuveges: Oh, I need to upload the input files to the google bucket.

DSuveges:
The update contains:
- All input files are on gs://
- The validation is turned on for all PPP sources.
- The path JSON schemas were updated across all rules.

Caveats/ still missing:
- As the input files for the PPP are not read from snakemake, but a config json, when data is read from gs://, process fails unless run on dataproc.
- As far as I know, the service account we are using doesn't have write access on gs://otar013-ppp/, we are working on with the backend team.

common/evidence.py
ireneisdoomed: @DSuveges created a ticket about this and it is scoped for the next release

tskir: Thank you! Since there is a ticket, I'm marking this thread as resolved

configuration.yaml
```diff
@@ -72,3 +72,15 @@ TargetSafety:
   toxcast: gs://otar001-core/TargetSafety/data_files/toxcast/ToxCast_2021-08-17.tsv
   aopwiki: gs://otar001-core/TargetSafety/data_files/aopwiki/aopwiki-2023-01-20.json
   outputBucket: gs://otar001-core/TargetSafety/json
+OT_CRISPR:
+  data_directory: gs://otar013-ppp/ot_crispr/input_data
+  config: gs://otar013-ppp/PPP-evidencie-configuration/ot_crispr_config.json
```
tskir, line 77: There's a typo: "evidencie" instead of "evidence". (The GS folder also has this typo, so it should be renamed both there and in the PR)

tskir (changes requested):
Everything looks good to me except for one typo as mentioned

configuration.yaml
DSuveges: Oh, I can't believe I'm making mistakes like this. Thanks for spotting!

tskir (approved):
Everything looks good to me now!

PR #158, opened 2023-04-20
Adding parsers to generate evidence based on publicly available CRISPR screens

DSuveges:
At this point the Crispr Brain dataset is ingested. 

Disease mappings are coming from this file: `https://gist.githubusercontent.com/DSuveges/218e1a5747b287423654ec16a0b37124/raw/2becccda79c9b7a42a139212e4ac3c247cf6611a/brain_crispr_studies.tsv`

* Calling the script:
```bash
python modules/crispr_screens.py \
    --crispr_brain_mapping brain_crispr_studies.tsv \
    --output crisprscreen.json.gz
```

The schema looks like this:
```json
{
  "studyId": "Glutamatergic Neuron-Survival-CRISPRi",
  "datatypeId": "sysbio",
  "datasourceId": "crispr_screen",
  "projectId": "crispr_brain",
  "studyOverview": "CRISPRi survival screen in human iPSC-derived glutamatergic neurons (standard medium).",
  "experiment": "WT CRISPRi-iPSCs were transduced with genome-wide CRISPRi-v2 sgRNA libraries. Two days after infection, the cells were selected for lentiviral integration using puromycin (1 g/mL) for 3 days as the cultures were expanded for the screens.",
  "cellType": "Glutamatergic Neuron",
  "diseaseFromSource": "essential genes / neurodegenerative disease",
  "diseaseFromSourceMappedId": "EFO_0005772",
  "contrast": "Survival",
  "targetFromSource": "SOD2",
  "pValue": 0.0029283011705045,
  "statisticalTestTail": "lower tail",
  "resourceScore": -24.05642773885583
}
```
**TODO**:
- [x] The snakemake file is not yet updated. 
- [x] The evidence definition is not yet added to the json schema.
- [ ] Mappings are not yet added to the curation repo.

### DIFF
--- a/modules/crispr_screens.py
+++ b/modules/crispr_screens.py
@@ -31,7 +31,7 @@ def main(spark: SparkSession, crispr_brain_mapping: str) -> DataFrame:
         crispr_screen_evidence_sets,
     )
 
-    logging.info(f"Total number of gene_burden evidence: {combined_evidence.count()}")
+    logging.info(f"Total number of CRISPR screen evidence: {combined_evidence.count()}")
     return combined_evidence
 
 


DSuveges:
I think the code should be reviewed before it become too complicated.

DSuveges:
- Also added to Snakemake
- JSON schema available. 

Generated evidence:
```json
{
  "studyId": "Glutamatergic Neuron-Survival-CRISPRi",
  "datatypeId": "affected_pathway",
  "datasourceId": "crispr_screen",
  "projectId": "crispr_brain",
  "crisprScreenLibrary": "Genome-wide CRISPRi-v2",
  "studyOverview": "CRISPRi survival screen in human iPSC-derived glutamatergic neurons (standard medium).",
  "cellType": "Glutamatergic Neuron",
  "literature": [
    "34031600"
  ],
  "diseaseFromSource": "essential genes / neurodegenerative disease",
  "diseaseFromSourceMappedId": "EFO_0005772",
  "contrast": "Survival",
  "targetFromSourceId": "SOD2",
  "resourceScore": 0.0029283011705045,
  "statisticalTestTail": "lower tail",
  "log2FoldChangeValue": -24.05642773885583
}
```

Validated against [this schema](https://raw.githubusercontent.com/opentargets/json_schema/40128495bb2bcf9efc4b7bb1a4fe84632835a524/schemas/disease_target_evidence.json) on a PR.

configuration.yaml
```diff
@@ -67,8 +68,8 @@ SysBio:
 TEP:
   outputBucket: gs://otar001-core/TEPs
 TargetSafety:
-  adverseEvents: https://raw.githubusercontent.com/opentargets/curation/23.02/target_safety/adverse_effects.tsv
-  safetyRisk: https://raw.githubusercontent.com/opentargets/curation/23.02/target_safety/safety_risks.tsv
+  adverseEvents: target_safety/adverse_effects.tsv
+  safetyRisk: target_safety/safety_risks.tsv
```
DSuveges, line 72: Having links to versioned repos were a source of errors. The root to the repo + version had to be moved to the globals.

tskir: I really like this, now it's much easier to migrate to newer version of curation data

configuration.yaml
```diff
@@ -84,3 +85,7 @@ Encore:
   data_directory: gs://otar013-ppp/encore/input
   config: gs://otar013-ppp/PPP-evidence-configuration/encore_config.json
   outputBucket: gs://otar013-ppp/encore
+CrisprScreens:
+  outputBucket: gs://otar000-evidence_input/Crispr_screens/json
+  # This line needs to be updated once the mapping is merged into the curation repo:
+  crispr_brain_mapping: https://gist.githubusercontent.com/DSuveges/218e1a5747b287423654ec16a0b37124/raw/2becccda79c9b7a42a139212e4ac3c247cf6611a/brain_crispr_studies.tsv
```
tskir, line 91: Is there something blocking us from merging this into the curation repo straight away? I'd really prefer not to merge a hardcoded gist link into the main branch here :sweat_smile:

modules/BrainCRISPR.py
```diff
@@ -0,0 +1,278 @@
...
+from pyspark import SparkFiles
+
+if TYPE_CHECKING:
+    from pyspark.sql import DataFrame, SparkSession
+
+
+class CRISPRBrain:
+    # CRISPR Brain API details. API key is required and also allows for checking for new release.
+    API_SERVER_URL = "https://crisprbrain.org"
+    API_CLIENT_ID = "92f263647c43525d3e4f181aa7e348f26e32129c0f827321d9261cc9765c56c0"
```
tskir, line 24: If they have an API key (vs. just having open access), isn't it supposed to be a secret? Is it OK to have it hardcoded here in the open?

modules/BrainCRISPR.py
```diff
@@ -0,0 +1,278 @@
...
+            disease_mapping_url (str): file containing the study to disease mapping
+        """
+        self.spark = spark
+
+        # Adding disease mapping file to spark context:
+        self.spark.sparkContext.addFile(disease_mapping_url)
+        self.disease_mapping_file = disease_mapping_url.split("/")[-1]
+
+    def __get_study_table(self: CRISPRBrain) -> DataFrame:
+        """Extract scree data from Crispr brain API.
```
tskir, line 55:
```suggestion
        """Extract screen data from Crispr brain API.
```

modules/BrainCRISPR.py
```diff
@@ -0,0 +1,278 @@
...
+        )
+
+    def create_crispr_brain_evidence(self: CRISPRBrain) -> None:
+        """"""
+        # Reading
+        crispr_studies = (
+            # Reading studies:
+            self.__get_study_table()
+            # Joining with resolved literature:
+            .join(self.__get_literature_mapping(), on="Reference Link", how="left")
```
tskir, line 220: I see a potential problem here: if in the future they add more publications, the lut returned by `__get_literature_mapping()` will not be able to map them all, and will just _silently_ continue. I'm not sure what is the best solution here though. But perhaps displaying at least a warning would be a good idea?

modules/BrainCRISPR.py
```diff
@@ -0,0 +1,278 @@
...
+                # Renaming fields:
+                f.col("Screen Name").alias("studyId"),
+                f.col("Libraries Screened").alias("crisprScreenLibrary"),
+                f.col("studySummary.title").alias("studyOverview"),
+                # Dropping experiment column for now:
+                # f.col("studySummary.experiment").alias("experiment"),
+                f.col("Cell Type").alias("cellType"),
+                f.when(f.col("Genotype") != "WT", f.col("Genotype"))
+                .otherwise(None)
+                .alias("geneticBackground"),
```
tskir, line 235: This is not strictly necessary, but I think it would be better to make this section slightly easier to read, maybe something like this:

```suggestion
                f.when(f.col("Genotype") != "WT", f.col("Genotype"))
                    .otherwise(None)
                    .alias("geneticBackground"),
```

modules/BrainCRISPR.py
```diff
@@ -0,0 +1,278 @@
...
+                "Phenotype",
+                "literature",
+            )
+            # Joining resolved literature:
+            .join(self.__get_disease_mapping(), on="studyId", how="inner")
```
tskir, line 240: Oof, this worries me even more than the literature thing I mentioned before. Here we have an inner join, which means that all evidence without a mapped disease will be just silently dropped.

Do you know many many traits are there in total in that dataset? Are they likely to be adding more, and how frequently? I'm just trying to estimate the scale of the potential problem

tskir (commented):
Very much like it overall, just a few minor suggestions, & two questions pertaining to curating literature/disease mappings and how are we planning on keeping those updated

modules/BrainCRISPR.py
DSuveges: This key is inside the [API wrapper code](https://pypi.org/project/crisprbrain/). Not sure if we can or cannot include in the code.... Not sure how to work around it.

modules/BrainCRISPR.py
DSuveges: Any upcoming release will *likely* break the API:
- The [API key](https://github.com/opentargets/evidence_datasource_parsers/blob/bbc2ac9a486d70c0150603513205638e6d522c91/modules/BrainCRISPR.py#L24) will likely changed.
- Also the [API version](https://github.com/opentargets/evidence_datasource_parsers/blob/bbc2ac9a486d70c0150603513205638e6d522c91/modules/BrainCRISPR.py#L71) should be different as well. 

These checks are also implemented into the official wrapper.

modules/BrainCRISPR.py
DSuveges: I don't know... as far as I can see there was no update. I can't even find release date on the webpage as well.

DSuveges: I can certainly work around the joins, handle/log the inconsistencies in the study table, then drop rows with missing values.

DSuveges:
I have update the parser by not just applying a filtering join, but getting some sort of feedback on the counts:

```txt
2023-05-16 14:54:03 INFO BrainCRISPR - __get_disease_mapping: Number of studies with disease mapping: 23
2023-05-16 14:54:04 INFO BrainCRISPR - __QC_studies: Number of studies without literature mapping: 5
2023-05-16 14:54:05 INFO BrainCRISPR - __QC_studies: Number of studies without disease mapping: 11
2023-05-16 14:54:05 INFO BrainCRISPR - __QC_studies: Number of studies with disease mapping: 23
2023-05-16 14:54:21 INFO crispr_screens - main: Total number of gene_burden evidence: 26885
```

In this case, we are informed that there are 23 studies with disease mapping and the same number of studies propagated towards the evidence. If something changes, the number of unmapped studies increase and/or we won't see as much mapped studies in the final dataset.

modules/BrainCRISPR.py
tskir: > I can certainly work around the joins, handle/log the inconsistencies in the study table, then drop rows with missing values.

Yes, I think this be good! I see that now this is implemented as a left join instead of inner (which I welcome). Should we add a `logging.warn()` call as well to report there are <this many> evidence strings without disease/literature mapping?

modules/BrainCRISPR.py
tskir: UPD: I now see below that the logging has already been implemented, now it's all good. Resolving this thread

tskir (approved):
I went through all discussion and can confirm there's nothing outstanding remaining. Everything looks good to me. And thank you for adding logging for the counts of evidence strings missing certain fields!

I'm approving now, but will leave the merge to you in case you wanted to do some final touches

PR #163, opened 2023-06-12
Update crispr parser

DSuveges:
making small modifications to accommodate studies with multiple filter criteria.


DSuveges:
This line:
```
        # Which end of the distribution are we looking? - "neg" or "pos"?
        side = filterColumn.split("|")[0]
```
Could be especially cryptic without explanation.

PR #166, opened 2023-08-10
P&Ds tables parser

ireneisdoomed:
The latest release of Probes&Drugs included all the data that we need to generate a chemicalProbes file with 2 files:
- A Probes Excel Dump - with all the information needed to parse the probes without setting the SQL db
- A Probes to ChEMBLID mappings table - that removes the dependency we had on our drug index

This version of the parser produces a `chemicalProbes.json` file that is based on the [latest P&D release](https://www.probes-drugs.org/news#48), passes validation, and is integrated in our Snakefile.

The latest data is available at `gs://otar001-core/ChemicalProbes/annotation/chemicalprobes-2023-08-10.json.gz`, with ~5000 probes defined for ~900 unique targets.


tskir (approved):
Thank you! Everything looks good to me

PR #172, opened 2023-10-03
New parser: Direct ingestion of GTEx V8 for preparation of baseline expression data

tskir:
Directly ingests GTEx V8 by-gene median expression values; applies three sets of metrics: Gini, HPA, Adatiss; outputs the results according to the [baseline expression schema](https://github.com/opentargets/json_schema/blob/master/schemas/baseline_expression.json).

Modified version of the Adatiss code is bundled together with the parser because the original contains a mathematical bug, which the author hasn't fixed since my suggestion in March this year (https://github.com/mwgrassgreen/AdaTiSS/issues/2)

As agreed with @DSuveges, even though most of our parsers use PySpark, this one is deliberately implemented in Pandas. This is because we have wide data as input, and Adatiss is also only able to process wide data. So the processing would look something like this:
* In PySpark: ingest wide → convert to long and process → convert back to wide to pass for Adatiss → read results from Adatiss and convert back to long → process and output.
* In pandas: ingest wide → process wide through Adatiss → pack the final results into the long output format.

Changes to schema and curation repository versions are introduced to ensure that you can reproduce running this code. Once the PR is approved, we can either do intermediate releases on both the curation and schema repositories; or keep the versions as "master" until we generate the data for the upcoming release.

The data output for this code can be found here: gs://otar000-evidence_input/BaselineExpression/json/baseline_expression-2023-10-03.json.gz


configuration.yaml
```diff
@@ -1,10 +1,15 @@
 global:
   logDir: gs://otar000-evidence_input/parser_logs
   cacheDir: cache_dir
-  schema: https://raw.githubusercontent.com/opentargets/json_schema/2.4.0
+  schema: https://raw.githubusercontent.com/opentargets/json_schema/master
```
DSuveges, line 4: This works and robust, however when preparing for release I think will be nicer to refer for a specific release from the json_schema repo.

configuration.yaml
```diff
@@ -1,10 +1,15 @@
...
   EFOVersion: v3.57.0
   cell_passport_file: https://cog.sanger.ac.uk/cmp/download/model_list_latest.csv.gz
-  curation_repo: https://raw.githubusercontent.com/opentargets/curation/23.09
+  curation_repo: https://raw.githubusercontent.com/opentargets/curation/tskir-gtex-v8
```
DSuveges, line 7: Also, this needs to be updated once we are preparing for release.

DSuveges (approved):
All looks good.

PR #178, opened 2024-01-15
feat: import_trait_mappings reads disease curation from configIl 2907

ireneisdoomed:
This PR includes:
- fixes to the problem reported in [#2907](https://github.com/opentargets/issues/issues/2907), where the disease mappings path was hardcoded to the 22.09 release. We now build the path from the project's config YAML
- Spark instance manipulation is no longer handled inside `import_trait_mappings`. It is passed as a parameter instead

This was an important bug to solve because we were not using the most updated version of the disease mappings file.
Luckily, **this error had no impact**. It was only used by the Genebass, and AZ parsers and the curation for their evidence hasn't changed since 22.09.

I've tested it locally with a subset of data and it worked.

### DIFF
--- a/common/evidence.py
+++ b/common/evidence.py
@@ -1,10 +1,13 @@
 import logging
 import os
 from psutil import virtual_memory
-import requests
-import tempfile
 import json
+import yaml
+import tempfile
+from typing import Dict
+
 import gcsfs
+import requests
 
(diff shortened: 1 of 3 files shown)


DSuveges (approved):
So much nicer!

PR #186, opened 2024-02-16
Feature: Updating Project score datasource

DSuveges:
As part of the effort described under [#3027](https://github.com/opentargets/issues/issues/3027)

What is happening:

- The new version of ProjectScore data is integrated into the pipeline.
- Parser renamed from `Crispr` to `PojectScore`.
- The class (`GenerateDiseaseCellLines`) that processes Cell passport cell annotation is now fetching tissue label to uberon mapping from a manually curated file instead of API calls.

!! Important !! The generated evidence is failing agains the current, fixed schema, however successfully validates against the schema on the master.


Snakefile
```diff
@@ -23,7 +23,7 @@ ALL_FILES = [
     ('chembl.json.gz', GS.remote(f"{config['ChEMBL']['outputBucket']}/chembl-{timeStamp}.json.gz")),
     ('clingen.json.gz', GS.remote(f"{config['ClinGen']['outputBucket']}/clingen-{timeStamp}.json.gz")),
     ('clingen-Gene-Disease-Summary.csv', GS.remote(f"{config['ClinGen']['inputBucket']}/clingen-Gene-Disease-Summary-{timeStamp}.csv")),
-    ('crispr.json.gz', GS.remote(f"{config['CRISPR']['outputBucket']}/crispr-{timeStamp}.json.gz")),
+    ('project_score.json.gz', GS.remote(f"{config['ProjectScore']['outputBucket']}/project_score-{timeStamp}.json.gz")),
```
ireneisdoomed, line 26: Did you generate evidence with all changes here?
Maybe it's a dumb question, but I don't see any file named `project_score` in the bucket.

common/evidence.py
```diff
@@ -1,19 +1,54 @@
...
+from psutil import virtual_memory
 from pyspark import SparkFiles
+from pyspark.conf import SparkConf
+from pyspark.sql import DataFrame, SparkSession
+
+if TYPE_CHECKING:
+    from pyspark.sql import Column, DataFrame
+
+
+def initialize_logger(name: str, log_file: Optional[str] = None) -> None:
```
ireneisdoomed, line 23: Thank you for this

Snakefile
```diff
@@ -200,25 +200,27 @@ rule clingen:                 # Process the Gene Validity Curations table from C
...
 
-rule crispr:                  # Process cancer therapeutic targets using CRISPR–Cas9 screens.
+rule projectScore:                  # Process cancer therapeutic targets using CRISPR–Cas9 screens.
     input:
-        evidenceFile = GS.remote(config['CRISPR']['inputAssociationsTable']),
-        descriptionsFile = GS.remote(config['CRISPR']['inputDescriptionsTable']),
-        cellTypesFile = GS.remote(config['CRISPR']['inputCellTypesTable'])
+        evidenceFile = GS.remote(config['ProjectScore']['geneScores']),
+        cellTypesFile = GS.remote(config['ProjectScore']['cellLinesTable']),
+        uberontoCellLineMapping = HTTP.remote(f"{config['global']['curation_repo']}/{config['Essentiality']['depmap_tissue_mapping']}"),
```
ireneisdoomed, line 207: Since you've added the logic to deal with the remote file, I'd suggest you move this from `input` (it downloads the data), to `params` (it just creates the environment variable)

common/evidence.py
```diff
@@ -103,87 +145,87 @@ def generate_map(self) -> None:
...
-    @staticmethod
-    @f.udf(
-        ArrayType(
-            StructType(
-                [
-                    StructField("name", StringType(), nullable=False),
-                    StructField("description", StringType(), nullable=False),
-                ]
+        # If the mapping file is a URL, download it and read it into a Spark dataframe.
+        if "http" in cell_line_to_uberon_mapping:
```
ireneisdoomed, line 199: I THINK using SparkFiles with a local filepath also worked, but I don't remember exactly. Did you test it didn't work before coding both scenarios?

common/evidence.py
```diff
@@ -103,87 +145,87 @@ def generate_map(self) -> None:
...
+                ),
+            )
+            .when(
+                status == "MSS",
+                f.struct(
+                    f.lit("MSS").alias("name"),
+                    f.lit("Microsatellite stable").alias("description"),
+                ),
+            )
+            .otherwise(f.lit(None))
```
ireneisdoomed, line 227: Redundant

common/evidence.py
```diff
@@ -238,23 +280,25 @@ def read_path(path: str, spark_instance) -> DataFrame:
         return spark_instance.read.option("recursiveFileLookup", "true").json(path)
 
     # A directory with Parquet files.
-    if parquet_files:
+    else:
```
ireneisdoomed, line 283: This wouldn't work if the folder contained CSVs.

modules/ProjectScore.py
```diff
@@ -0,0 +1,186 @@
...
+from pyspark.sql import types as t
+
+from common.evidence import (
+    GenerateDiseaseCellLines,
+    initialize_logger,
+    initialize_sparksession,
+    write_evidence_strings,
+)
+
+# This parser is specific for the second release of Porject Score, so the publication identifier is hardcoded
```
ireneisdoomed, line 16:
```suggestion
# This parser is specific for the second release of Project Score, so the publication identifier is hardcoded
```

modules/ProjectScore.py
```diff
@@ -0,0 +1,186 @@
...
+    """Generate evidence strings for Project Score v2.
+
+    Args:
+        project_score_cell_lines (DataFrame): DataFrame containing cell line annotations
+        project_score_evidence (DataFrame): DataFrame containing evidence data
+
+    Returns:
+        DataFrame: DataFrame containing evidence strings
+    """
+    pmid = "38215750"
```
ireneisdoomed, line 32:
```suggestion
    PMID = "38215750"
```

modules/ProjectScore.py
```diff
@@ -0,0 +1,186 @@
...
+    return (
+        project_score_evidence.select(
+            f.col("targetSymbol").alias("targetFromSource"),
+            f.col("diseaseName").alias("diseaseFromSource"),
+            f.col("diseaseId").alias("diseaseFromSourceMappedId"),
+            f.col("PRIORITY").cast(t.FloatType()).alias("resourceScore"),
+            f.col("targetId").alias("targetFromSourceId"),
+            f.array(f.lit(pmid)).alias("literature"),
```
ireneisdoomed, line 40:
```suggestion
            f.array(f.lit(PMID)).alias("literature"),
```

modules/ProjectScore.py
```diff
@@ -0,0 +1,186 @@
...
+
+from common.evidence import (
+    GenerateDiseaseCellLines,
+    initialize_logger,
+    initialize_sparksession,
+    write_evidence_strings,
+)
+
+# This parser is specific for the second release of Porject Score, so the publication identifier is hardcoded
+PMID = "38215750"
```
ireneisdoomed, line 17: This is inside `generate_project_score_evidence`
```suggestion
```

modules/ProjectScore.py
```diff
@@ -0,0 +1,186 @@
...
+            f.array(f.lit(pmid)).alias("literature"),
+            f.lit("crispr").alias("datasourceId"),
+            f.lit("affected_pathway").alias("datatypeId"),
+            f.lower(f.col("cancerType")).alias("cancerType"),
+        )
+        .join(project_score_cell_lines, on="cancerType", how="left")
+        # Dropping the pancancer genes:
+        .filter(f.col("cancerType") != "pancancer")
+        # Cleaning table:
+        .drop("cancerType")
```
ireneisdoomed, line 49: Incredibly minor, but if you do the select here you can omit this drop

modules/ProjectScore.py
```diff
@@ -0,0 +1,186 @@
...
+
+    # Extract disease cell line data from cell passport file:
+    cell_passport_data = GenerateDiseaseCellLines(
+        spark, cell_passport_file, cell_line_to_uberon_mapping
+    )
+
+    passport_disease_cell_lines = cell_passport_data.generate_disease_cell_lines()
+
+    # Joining disease cell-lines dataframe for Project Score:
+    disease_cell_lines = (
```
ireneisdoomed, line 81: I'd suggest to extract this logic into a function

ireneisdoomed (commented):
Thank you! 
The data structure has changes, so i think that justifies storing the data in a separate folder. I don't have major comments.

The only thing is that this code will generate a file with the pattern `project_score-{date}.json.gz`. That's not the filename we have at the moment, everything in the bucket is named `crispr`.

Approving these suggestions will mean a change of filename, which will require an update in PIS. If we are going to involve other teams, I'd suggest thinking about renaming the datasource entirely to `project_score`

DSuveges:
> The data structure has changes, so i think that justifies storing the data in a separate folder. 

For consistency, I moved the old files into a separate folder and the v2 folder moved into the data folder:

```
gs://otar000-evidence_input/CRISPR/data_files/
gs://otar000-evidence_input/CRISPR/data_files/project_score_v1/
gs://otar000-evidence_input/CRISPR/data_files/project_score_v2/
```

> The only thing is that this code will generate a file with the pattern `project_score-{date}.json.gz`. That's not the filename we have at the moment, everything in the bucket is named `crispr`.
>
> Approving these suggestions will mean a change of filename, which will require an update in PIS. If we are going to involve other teams, I'd suggest thinking about renaming the datasource entirely to `project_score`

Yes, precisely,  I did used the old filename, as I didn't want to complicate stuff downstream, and my original intention was to rename this data source completely in the whole infrastructure. 'crispr' doesn't make any sense. It was out of scope for this release, but if capacity is given, for the next release maybe. It's a refactor with not much value.

modules/ProjectScore.py
DSuveges: Oh, yes.

Snakefile
DSuveges: Yes, see my other reply.

common/evidence.py
DSuveges: I have tested it with and without specifying the logfile it seems it works. once we are touching other parsers, logging can be updated.

common/evidence.py
DSuveges: I have tried both remote and local files, both work.

common/evidence.py
DSuveges: I like when things are explict, but fine. :D

common/evidence.py
DSuveges: If the path only contains csv-s, the above assertion would have already failed:

```python
assert (
    json_files or parquet_files
), f"The provided directory {path} contains neither JSON nor Parquet."
```

However if a folder contains a mix of csv and parquet file it would indeed fail, but not because of the condition, but because of this in line 284:

```python
        return spark_instance.read.parquet(path)
```

More over this would also fail:

```python
    if json_files:
        return spark_instance.read.option("recursiveFileLookup", "true").json(path)
```

I think it is safe to assume if a directory is provided, it is either a propery partitioned json or parquet dataset. If it's a partitioned tsv, it won't fly.

modules/ProjectScore.py
DSuveges: Interesting idea. I moved all cell-line related logic into a separate function things become neater. Thanks.

modules/ProjectScore.py
DSuveges: I have removed this one, and leaved the one outside the function. In the top of the file.

modules/ProjectScore.py
ireneisdoomed: Thanks!

modules/ProjectScore.py
```diff
@@ -49,6 +50,45 @@ def generate_project_score_evidence(
...
+        cell_line_data.select(
+            f.lower(f.col("CANCER_TYPE")).alias("cancerType"),
+            f.col("CMP_ID").alias("id"),
+        )
+        .join(passport_disease_cell_lines, on="id", how="right")
+        .groupBy("cancerType")
+        .agg(f.collect_set("diseaseCellLine").alias("diseaseCellLines"))
+    )
+    # Get report:
+    logger = logging.getLogger(__name__)
```
ireneisdoomed, line 78: I'd suggest using the same logger we set in `main`, instead of initialising a new one.
To do that, I see 3 easy options:
1.  either return `disease_cell_lines` here and move the reporting logic to main (my preferred option because of separation of concerns)
2. pass the logger as an argument of `get_disease_cell_lines`
3. in main, make `logger` a global variable
```
global logger
logger = logging.getLogger(__name__)
```

ireneisdoomed:
Thank you for the explanation and your changes! I also vote for renaming the source as a whole.
Just 2 minor comments

modules/ProjectScore.py
DSuveges: In my understanding, that statement instead of initializing the logger, it retrieves the logger that was already been formatted for the module, as `__name__` is given, which is module-wide identifier. Also used identically in the main function as well (line 111). I'm not clear on the details, but as seen on the logs, messages from this function is definitely formatted consistently (implies that they are the same loggers), also true if `--log_file` is specified:

```
2024-03-12 12:50:00 INFO ProjectScore - main: Running Project Score parser with parameters:
2024-03-12 12:50:00 INFO ProjectScore - main: Evidence file: /Users/dsuveges/Downloads/project_score_v2/mapped_diseases.tsv
2024-03-12 12:50:00 INFO ProjectScore - main: Cell types file: /Users/dsuveges/Downloads/project_score_v2/cell_types.tsv
2024-03-12 12:50:00 INFO ProjectScore - main: Cell passport file: /Users/dsuveges/Downloads/model_list_20240110.csv
2024-03-12 12:50:00 INFO ProjectScore - main: Cell line to uberon mapping: depmap_uberon_mapping.csv
2024-03-12 12:50:00 INFO ProjectScore - main: Output file: projectscore_v2b.json.gz
2024-03-12 12:50:05 INFO ProjectScore - get_disease_cell_lines: All cancer types are in the cell line file.
2024-03-12 12:50:06 INFO ProjectScore - main: Number of evidence strings: 517
2024-03-12 12:50:07 INFO ProjectScore - main: Number of targets in evidence data: 302
2024-03-12 12:50:07 INFO ProjectScore - main: Number of diseases in evidence data: 26
2024-03-12 12:50:07 INFO ProjectScore - main: Writing evidence strings to file...
```

Snakefile
DSuveges: However this is true for all remote files used in the pipeline.

ireneisdoomed (approved):
Thanks!

PR #192, opened 2024-05-02
feat: updating validation lab parser for the new data

DSuveges:
The following updates were implemented:

- The script creates evidence following the schema defined in [this PR](https://github.com/opentargets/json_schema/pull/195)
- The script uses the updated validation lab configuration updated [here](https://github.com/opentargets/PPP-evidence-configuration/blob/master/ValidationLab_config.json)
- The script processes the newly generated data files by the validation lab. (mock data here: `gs://otar013-ppp/validation_lab/input_data/2024.04`)
- The script still relies on the Sanger modell passport data.

Calling the script:

```bash
python partner_preview_scripts/ValidationLab.py \
    --output_file output.json.gz \
    --parameter_file ../PPP-evidence-configuration/ValidationLab_config.json \
    --cell_passport_file  ~/project_data/validation_lab/model_list_latest.csv.gz \
    --data_folder ~/project_data/validation_lab/
```

The resulting output validates against the above cited evidence schema.

### DIFF
--- a/partner_preview_scripts/ValidationLab.py
+++ b/partner_preview_scripts/ValidationLab.py
@@ -3,10 +3,10 @@
 
 - The parser uses results from various screening experiments.
 - Input files are defined in the ValidationLab_config.json configuration in the PPP-evidence-configuration repo
-- Files:
+- Data files:
     - Cell line data with biomarker annotation
     - Result file with the measured values.
-    - Hypothesis files with the binary flags for each tested biomarker in the context of the tested gene.
+    - Sanger Model Passport data.
 """
 
 import argparse
@@ -17,8 +17,7 @@ from functools import reduce
 
 import pyspark.sql.functions as f
 import pyspark.sql.types as t
-from pyspark.sql import SparkSession
-from pyspark.sql.dataframe import DataFrame
+from pyspark.sql import Column, DataFrame, SparkSession
 
 from common.evidence import (
     initialize_logger,
@@ -29,6 +28,22 @@ from common.evidence import (
 
 
 class BiomarkerParser:
+    """This class is responsible for parsing biomarker data.
+
+    There are a handful of accepted biomarkers listed as dictionary keys. These values are expected to be
+    column names in the biomarker input table. Then the column values are mapped to a standardised format.
+
+    The output expected to collate all the biomarkers into a single column for each cell line. With the
+    follwoing schema:
+
+    root
+    |-- cellName: string (nullable = true)
+    |-- biomarkers: array (nullable = false)
+    |    |-- element: struct (containsNull = false)
+    |    |    |-- name: string (nullable = false)
+    |    |    |-- description: string (nullable = false)
+    """
+
     BIOMARKERMAPS = {
         "PAN": {
             "direct_mapping": {
@@ -107,43 +122,47 @@ class BiomarkerParser:
     }
 
     @staticmethod
-    @f.udf(
-        t.StructType(
-            [
-                t.StructField("name", t.StringType(), False),
-                t.StructField("description", t.StringType(), False),
-            ]
+    def _get_biomarker_wrapper(biomarker_map: dict):
+        @f.udf(
+            t.StructType(
+                [
+                    t.StructField("name", t.StringType(), False),
+                    t.StructField("description", t.StringType(), False),
+                ]
+            )
         )
-    )
-    def _get_biomarker(column_name, biomarker, BIOMARKERMAPS):
-        """This function returns a struct with the biomarker name and description."""
+        def wrapped_function(column_name: Column, biomarker: Column):
+            """This function returns a struct with the biomarker name and description."""
 
-        # If the biomarker has a direct mapping:
-        if "direct_mapping" in BIOMARKERMAPS[column_name]:
-            try:
-                return BIOMARKERMAPS[column_name]["direct_mapping"][biomarker]
-            except KeyError:
+            # If the biomarker value has a direct mapping:
+            if "direct_mapping" in biomarker_map[column_name]:
+                try:
+                    return biomarker_map[column_name]["direct_mapping"][biomarker]
+                except KeyError:
+                    logging.warning(
+                        f"Could not find direct mapping for {column_name}:{biomarker}"
+                    )
+                    return None
+
+            # If the value needs to be parsed:
+            if biomarker == "wt":
+                return {
+                    "name": biomarker_map[column_name]["name"] + biomarker,
+                    "description": biomarker_map[column_name]["description"]
+                    + "wild type",
+                }
+            elif biomarker == "mut":
+                return {
+                    "name": biomarker_map[column_name]["name"] + biomarker,
+                    "description": biomarker_map[column_name]["description"] + "mutant",
+                }
+            else:
                 logging.warning(
-                    f"Could not find direct mapping for {column_name}:{biomarker}"
+                    f"Could not find direct mapping for {column_name}: {biomarker}"
                 )
                 return None
 
-        # If the value needs to be parsed:
-        if biomarker == "wt":
-            return {
-                "name": BIOMARKERMAPS[column_name]["name"] + biomarker,
-                "description": BIOMARKERMAPS[column_name]["description"] + "wild type",
-            }
-        elif biomarker == "mut":
-            return {
-                "name": BIOMARKERMAPS[column_name]["name"] + biomarker,
-                "description": BIOMARKERMAPS[column_name]["description"] + "mutant",
-            }
-        else:
-            logging.warning(
-                f"Could not find direct mapping for {column_name}: {biomarker}"
-            )
-            return None
+        return wrapped_function
 
     @classmethod
     def get_biomarkers(cls, biomarker_df):
@@ -160,8 +179,9 @@ class BiomarkerParser:
                 # Function to process biomarker:
                 lambda biomarker: (
                     biomarker,
-                    cls._get_biomarker(
-                        f.lit(biomarker), f.col(biomarker), cls.BIOMARKERMAPS
+                    cls._get_biomarker_wrapper(cls.BIOMARKERMAPS)(
+                        f.lit(biomarker),
+                        f.col(biomarker),
                     ),
                 ),
                 # Iterator to apply the function over:
@@ -199,7 +219,7 @@ class ValidationLabEvidenceParser:
 
     @staticmethod
     def _get_disease_cell_lines(model_passport_df: DataFrame) -> DataFrame:
-        return (
+        x = (
             # The following option is required to correctly parse CSV records which contain newline characters.
             model_passport_df.select(
                 f.col("model_name").alias("cellLineName"),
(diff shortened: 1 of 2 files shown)


partner_preview_scripts/ValidationLab.py
```diff
@@ -3,432 +3,472 @@
...
+        biomarkers_in_data = [
+            biomarker
+            for biomarker in cls.BIOMARKERMAPS.keys()
+            if biomarker in biomarker_df.columns
+        ]
 
-        # The first column is the gene name, the rest are the hypotheses:
-        hypothesis_columns = hypothesis_df.columns[1:]
+        # Applying the full map on the dataframe one-by-one:
+        processed_biomarkers = reduce(
```
ireneisdoomed, line 183: map+reduce are too complex for my brain, but if the intention is to create a new column per biomarker, could you just iterate over the list of those biomarkers present in the data? Sth like
```
  for biomarker in biomarkers_to_process:
    processed_biomarker = cls._get_biomarker_wrapper(biomarker_maps)(
        f.lit(biomarker), f.col(biomarker)
    )
    biomarker_df = biomarker_df.withColumn(biomarker, processed_biomarker)
```

partner_preview_scripts/ValidationLab.py
```diff
@@ -3,432 +3,472 @@
...
+                ),
+                # Iterator to apply the function over:
+                biomarkers_in_data,
+            ),
+            biomarker_df,
+        )
 
-        unpivot_expression = f"""stack({len(hypothesis_columns)}, {", ".join([f"'{x}', `{x}`" for x in hypothesis_columns])} ) as (hypothesis, {call})"""
+        # The biomarker columns are unstacked into one single 'biomarkers' column:
+        biomarker_unstack = f"""stack({len(biomarkers_in_data)}, {", ".join([f"'{x}', {x}" for x in biomarkers_in_data])}) as (biomarker_name, biomarker)"""
```
ireneisdoomed, line 201: Is a custom unstacking fucntion necessary or could we use `unpivot`? https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.unpivot.html

partner_preview_scripts/ValidationLab.py
```diff
@@ -3,432 +3,472 @@
...
+@dataclass
+class ValidationLabEvidenceParser:
+    projects: list
+    assays: list
+    sharedParemeters: dict
+
+    @staticmethod
+    def _get_disease_cell_lines(model_passport_df: DataFrame) -> DataFrame:
+        return (
+            # The following option is required to correctly parse CSV records which contain newline characters.
```
ireneisdoomed, line 230: This comment doesn't belong here
```suggestion
```

partner_preview_scripts/ValidationLab.py
```diff
@@ -3,432 +3,472 @@
...
+
+        # attributes added as column:
+        columns_to_add = [
+            "diseaseFromSourceMappedId",
+            "diseaseFromSource",
+            "tissueId",
+            "studyOverview",
+            "primaryProjectId",
+            "releaseVersion",
+            "releaseDate",
```
ireneisdoomed, line 357: The config says the data is from 2022, is this date up to date?

partner_preview_scripts/ValidationLab.py
```diff
@@ -3,432 +3,472 @@
...
-    # Reading and processing hypothesis data:
-    hypothesis_generator = ParseHypotheses(spark)
-    validation_hypotheses_df = hypothesis_generator.parse_hypotheses(
-        expectedFile=hypothesis_expected_file, observedFile=hypothesis_observed_file
-    )
+    @staticmethod
+    def _parser_assay_object(
+        raw_evidence_df: DataFrame, assays: list, spark: SparkSession
+    ) -> DataFrame:
+        """Organise experimental data into the right shape."""
```
ireneisdoomed, line 409: A more concrete description would be nice

partner_preview_scripts/ValidationLab.py
```diff
@@ -3,432 +3,472 @@
...
+                [
+                    # Generate evidence from a given project:
+                    ValidationLabProjectParser(**project).parse_evidence(
+                        spark, input_path, self.assays
+                    )
+                    for project in self.projects
+                    if not project["excludeStudy"]
+                ],
+            )
+            # Adding shared columns:
```
ireneisdoomed, line 264: You could move this whole select block to the next one

partner_preview_scripts/ValidationLab.py
```diff
@@ -3,432 +3,472 @@
...
+            .select(
+                # Columns from the raw evidence file:
+                "targetFromSourceId",
+                "cellLineName",
+                "assessment",
+                "resourceScore",
+                "primaryProjectHit",
+                # Column from the assay parser:
+                "assays",
+                # Columns from the project metadata:
```
ireneisdoomed, line 373: Aren't these already added later in the `generate_evidece` function?

partner_preview_scripts/ValidationLab.py
```diff
@@ -3,432 +3,472 @@
...
+            validation_input_file, sep="\t", header=True
         )
-        # Filter out all null biomarkers:
-        .filter(
-            (f.col("biomarkers").isNotNull())
-            &
-            # Following the request of the validation lab, we are removing CRIS biomarker annotation:
-            (f.col("biomarker_name") != "CRIS_subtype")
+        # First round of processing:
+        processed_experiment = self._process_raw_evidence(raw_experiment_data, assays)
```
ireneisdoomed, line 338: Maybe this distinction between `_process_raw_evidence` and `parse_evidence` is too artificial? I can see that `_process_raw_evidence` are just light transformations, but I don't know if the concern is separated enough to be in an indepedent function.

partner_preview_scripts/ValidationLab.py
```diff
@@ -3,432 +3,472 @@
...
+    ) -> DataFrame:
+        """Organise experimental data into the right shape."""
+        # Generate unpivot expression - not all assays are present in the data, some project might done with fewer assays:
+        assay_names = [
+            assay["shortName"]
+            for assay in assays
+            if assay["shortName"] in raw_evidence_df.columns
+        ]
+
+        unpivot_expression = f"""stack({len(assay_names)}, {', '.join([f"'{assay}', `{assay}`" for assay in assay_names])}) as (shortName, isHit)"""
```
ireneisdoomed, line 417: I think this is the same function as the unstacking one. I wonder the same, whether we can use the native `unpivot`

ireneisdoomed (commented):
Thank you! And sorry for the late review. This module is so flexible and modular to adjust for multiple assays that is really complex to follow. I hope we can process data from more experiments in the future and take advantage of it.

I haven't found anything logic related that could lead to a bug. My comments are mostly to increase readability

partner_preview_scripts/ValidationLab.py
DSuveges: When I learnt how to do unpivot, this functionality was not around. Also the linked docs says: `New in version 3.4.0.` so not sure if we can rely on this.

partner_preview_scripts/ValidationLab.py
DSuveges: Yes, indeed.

partner_preview_scripts/ValidationLab.py
DSuveges: Oh, I forget to do the update.

partner_preview_scripts/ValidationLab.py
DSuveges: Probably not. I was kind of struggling finding the right level of modularity. I'll take a look at how it could be made sensible. I just really don't like to write long functions (not that I was writing tests).

partner_preview_scripts/ValidationLab.py
```diff
@@ -3,432 +3,472 @@
...
-    )
-
-    # Applying the full map on the dataframe one-by-one:
-    biomarkers = reduce(
-        lambda DF, value: DF.withColumn(*value), expressions, validation_lab_cell_lines
-    )
+        return (
+            # Combining evidence from all validation projects
+            reduce(
+                lambda df1, df2: df1.unionByName(df1),
```
DSuveges, line 254: haha, there's something wrong here. :D

partner_preview_scripts/ValidationLab.py
DSuveges: I'm dropping this function, also quite artificial unit of logic

partner_preview_scripts/ValidationLab.py
```diff
@@ -3,432 +3,472 @@
...
+        },
+        "APC_status": {
+            "description": "APC mutation status: ",
+            "name": "APC-",
+        },
+        "BRAF_status": {"description": "BRAF mutation status: ", "name": "BRAF-"},
+    }
+
+    @staticmethod
+    def _get_biomarker_wrapper(biomarker_map: dict) -> Callable:
```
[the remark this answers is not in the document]
DSuveges: I did this I think because I couldn't pass self into udf. But I have to double check.

partner_preview_scripts/ValidationLab.py
DSuveges: fine.

partner_preview_scripts/ValidationLab.py
DSuveges: Also updated the ppp config files under `gs://otar013-ppp/PPP-evidence-configuration/`

partner_preview_scripts/ValidationLab.py
DSuveges: No, these are project specific constants not datasource specific ones. If I got your question right.

partner_preview_scripts/ValidationLab.py
DSuveges: Once we decided to bump pyspark version we can review the repo for other occurrences of this pattern.

DSuveges:
Thanks for review @ireneisdoomed ! Really helpful. I have addressed most of them, however there are a few, where I was unsure (eg. the udf one).

partner_preview_scripts/ValidationLab.py
ireneisdoomed: Good catch!

partner_preview_scripts/ValidationLab.py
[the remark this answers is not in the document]
ireneisdoomed: This thread says UDFs need to be static methods, because you only want to serialise a single function, not the function of an instance.
It could be as easy as defining the UDF as an static function like so.
I'd suggest trying, but of course it is not a blocker
```
@staticmethod
@f.udf(
            t.StructType(
                [
                    t.StructField("name", t.StringType(), False),
                    t.StructField("description", t.StringType(), False),
                ]
            )
        )
        def _get_biomarker_wrapper(column_name: Column, biomarker: Column) -> dict | None:
        ...
```

ireneisdoomed (approved):
Ty for the changes! If the output is good, all good from my side

partner_preview_scripts/ValidationLab.py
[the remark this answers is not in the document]
DSuveges: Yes, exactly, but in that case you need to bake the mapping dictionary into the udf. I wanted to avoid that.

PR #204, opened 2024-07-24
fix: change variant field to geneticVariation in cancerBiomarkers parser

vivienho:
The integration of variant entity in the Platform required to change the name of the a field in the biomarker evidence schema that should be reserved for the variant index.


DSuveges (approved):
All looks good. Script runs, the resulting dataset has the expected modification.

PR #205, opened 2024-07-26
fix: update ValidationLab parser to split assessment field by newline

vivienho:
Assessment field is also changed to plural (assessments) at relevant locations to reflect change to list of strings. These changes are to make the code compatible with downstream considerations.


DSuveges (approved):
All makes sense.

PR #217, opened 2024-11-13
Changes made to enable running parsers for 24.12 release

vivienho:
This PR updates the schema, curation, and EFO versions for the 24.12 release.

scipy (needed for the encore parser) was added to the requirements.

Changes were made to the IMPC, encore, and validation lab parsers.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -527,7 +527,8 @@ rule encore:                  # Generating PPP evidence for ENCORE
     params:
         data_folder = config['Encore']['data_directory'],
         config = config['Encore']['config'],
-        schema = f"{config['global']['schema']}/schemas/disease_target_evidence.json"
+        schema = f"{config['global']['schema']}/schemas/disease_target_evidence.json",
+	cell_line_to_uberon_mapping = f"{config['global']['curation_repo']}/{config['Essentiality']['depmap_tissue_mapping']}"
     output:
         'encore.json.gz'
     input:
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -12,6 +12,7 @@ pronto==2.5.3
 pyspark==3.3.1
 requests==2.32.0
 retry2==0.9.5
+scipy==1.10.1
```
DSuveges, line 15: Well spotted, we haven't updated the environment for a while.

partner_preview_scripts/encore_parser.py
```diff
@@ -514,13 +514,13 @@ def parse_experiment(self, parameters: dict) -> DataFrame:
...
 
-def main(outputFile: str, parameters: dict, cellPassportFile: str) -> None:
+def main(outputFile: str, parameters: dict, cellPassportFile: str, cellLineToUberonMapping: str) -> None:
 
     # Initialize spark session
     spark = initialize_sparksession()
 
     # Opening and parsing the cell passport data from Sanger:
-    cell_passport_df = GenerateDiseaseCellLines(cellPassportFile, spark).get_mapping()
+    cell_passport_df = GenerateDiseaseCellLines(spark, cellPassportFile, cellLineToUberonMapping).generate_disease_cell_lines()
```
DSuveges, line 523: The reason the encore parser is in such a bad shape is that the overall sentiment about encore data itself is extremely bad. Everyone assumes that the data is not good, and new analysis hopefully fixes these issues. As we have been waiting for ~2 years for the new release, and two contact people have left the encore team since, we have a good reason to assume, that the data we'll eventually get would be completely different from what we have now, so the entire parser will likely be rewritten. This project has a history and you are lucky that you are not part of this history.

Snakefile
DSuveges, line 1: Could you please comment out the ppp specific parsers in rows 52-53. These datasets are quite ad-hoc, and they usually quite rarely updated and in those cases we mostly need to touch the parsers. If they are not part of the general release process, it is fine.

DSuveges (approved):
Thank you for your effort to clean up the PPP madness. :D

PR #226, opened 2025-02-11
feat: use uv to install project

ireneisdoomed:


ireneisdoomed:
This is halted. These changes update the pyspark version which causes some parsers to fail. Will be reviewed after the release

DSuveges:
Thanks @ireneisdoomed for the work so far! I would argue that there's no need to distinguish between PPP and non-PPP parsers. I would just move the files to the same modules folder. Also, I know this is very marginal, but is `modules` the most appropriate name for this folder? (I'm just asking, because if we are already touching stuff, we can re-shuffle, however we want)

PR #228, opened 2025-02-19
build: upgrade yarl to fix environment conflict

ireneisdoomed:
This is a minor fix. The PGx evidence was failing because the environment was not properly set up (OpenAI was failing with [this error](https://community.openai.com/t/error-with-openai-1-56-0-client-init-got-an-unexpected-keyword-argument-proxies/1040332/2)).

I have fixed the environment so that `pip3 install -r requirements-frozen.txt` correctly sets up all dependencies. The changes in the pgx are only minor.


vivienho (approved):
All looks good!

PR #229, opened 2025-02-19
fix: fix orphanet

vivienho:
The orphanet parser was failing because of a `None`. This PR fixes the problem.


ireneisdoomed (changes requested):
Thank you for looking into it!!!
Looking more closely, I think this solution wouldn't cause the gene ID parsing to fail:
```
ensembl_gene_id = [
    xref.find('Reference').text
    for xref in gene.find('ExternalReferenceList')
    if xref.find('Reference') is not None and 'ENSG' in xref.find('Reference').text # now this wouldn't fail
]

if ensembl_gene_id:
    evidence['targetFromSourceId'] = ensembl_gene_id[0]
orphanet_disorders.append(evidence)
```

Can you give it a go and see if it runs? I think it looks better if we omit the try/except block

vivienho:
@ireneisdoomed The run failed :/ I think it's because checking the value of `xref.find('Reference')` is already causing the error

ireneisdoomed:
Ok, never mind. Thanks for checking!!!

ireneisdoomed (dismissed):

PR #231, opened 2025-03-31
feat(burden): extract evidence from cvdi human disease portal

ireneisdoomed:
TBD


DSuveges (approved):
Thank you @ireneisdoomed for curating the publication and writing the parser for yet an other gene burden dataset!

Very far fetched idea is coming: I'm wondering if the harmonisation of all these resources should be in the repo at all... I mean the source csv,xlsx etc files can be processed and harmonised to a shared schema only once. I assume probably the disease assignment (or maybe not even that) is the only process that needs to be re-run at each release.

PR #234, opened 2025-05-21
feat: prepare 25.06 release

ireneisdoomed:
This PR includes:
- Changes to the Gene2Phenotype module and configuration paths to adapt to their new schema of data.
- Update configuration to include latest data, and EFO version (3.77)
- Downgrade of the `h11` due to incompatibility with another library

This branch was the one I used to generate all evidence on the VM, with the exception of Gene2Phenotype which I run locally


DSuveges (approved):
Thank you so much for the update! All looks nice and make sense.

PR #237, opened 2025-06-11
fix: make join on diseaseFromSource null safe

vivienho:
I noticed that the Panelapp data always had a `null` `diseaseFromSourceMappedId` when the `diseaseFromSource` column is `null`. This is not because of a problem with mapping using the `diseaseFromSourceId`, but because the join on `diseaseFromSource` is not null safe, causing us to lose these mappings. This PR fixes the problem.


vivienho:
Out of 209 evidence rows with `null` `diseaseFromSource`, we now have `diseaseFromSourceMappedId` from 167 of them. With explosion due to multiple mappings, this now gives us 206 additional evidence rows. 42 evidence remain unmapped.

PR #244, opened 2025-08-12
fix(target safety): add evidence type to all sources

ireneisdoomed:
This PR closes https://github.com/opentargets/issues/issues/3643

The intention here is to add metadata to the target safety annotation so that users can filter based on the evidence provenance, as suggested by [this user ](https://community.opentargets.org/t/clinical-vs-pre-clinical-safety-liability-data/1613/2)

While fixing the study type for `Lamore` and `Force` evidence, I realised that other sources were also lacking the information because the `studies` field was set to null. We usually use `studies` to aggregate several evidence that inform about the same target/liability in a given source (think about the ToxCast or PharmGKB evidence [here](https://platform.opentargets.org/target/ENSG00000006638)). This was missing sometimes when the evidence came from literature curation.

After revision of the original publications to set a evidence type, this is what I finally came up with:
```
+----------------------------+---------------------+-----+
|array_distinct(studies.type)|datasource           |count|
+----------------------------+---------------------+-----+
|[clinical]                  |PharmGKB             |2159 |
|[preclinical]               |Lynch et al. (2017)  |1341 |
|[preclinical]               |Bowes et al. (2012)  |313  |
|[clinical]                  |Urban et al. (2012)  |254  |
|[cell-based]                |AOP-Wiki             |227  |
|[cell-based]                |ToxCast              |212  |
|[clinical]                  |Brennan et al. (2024)|194  |
|[biochemical]               |ToxCast              |167  |
|[preclinical]               |Force et al. (2011)  |47   |
|[cell-based]                |Lamore et al. (2017) |30   |
|[preclinical]               |Brennan et al. (2024)|16   |
|[null]                      |ToxCast              |1    | 
+----------------------------+---------------------+-----+
```

Preclinical when the evidence comes from animal studies, cell-based when it comes from assays that measure cellular responses, biochemical for assays measuring biochemical effects, and finally clinical when the information comes from directly measuring the effect in patients.

After these changes, only one evidence will be lacking this information, and this is coming directly from the raw data.


DSuveges (approved):
Thank you. As I understand, there's no schema change, as the `studyType` column is temporary and renamed to `type`, which was already in the schema.

PR #247, opened 2025-09-09
fix: incorrect handling of studies with multiple filter columns

DSuveges:
As https://github.com/opentargets/issues/issues/3347 explains, the parser was not handling well those situations where a list of filter columns were defined (eg. when both side of the distribution was considered).

### DIFF
--- a/partner_preview_scripts/ot_crispr.py
+++ b/partner_preview_scripts/ot_crispr.py
@@ -243,7 +243,7 @@ class OTAR_CRISPR_evidence_generator:
                 # Extract log2Fold change value based on where the hit is coming from:
                 f.when(f.col("sourceLabel").contains("pos"), f.col("pos|lfc"))
                 .when(f.col("sourceLabel").contains("neg"), f.col("neg|lfc"))
-                .otherwise(None).alias("log2FoldChangeValue"),
+                .otherwise(None).cast(t.FloatType()).alias("log2FoldChangeValue"),
                 
                 # Extract which tail of distribution the hit is coming from:
                 f.when(f.col("sourceLabel").contains("pos"), f.lit("upper tail")).when(
--- a/partner_preview_scripts/ot_crispr.py
+++ b/partner_preview_scripts/ot_crispr.py
@@ -14,6 +14,7 @@ import argparse
 import logging
 from functools import reduce
 from typing import Optional
+import operator
 
 from pyspark.sql import Column, DataFrame, Row, SparkSession
 from pyspark.sql import functions as f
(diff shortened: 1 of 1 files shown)


partner_preview_scripts/ot_crispr.py
```diff
@@ -241,7 +276,7 @@ def _read_and_filter_mageck_file(
             raise ValueError(f"Unrecognized label separator in {raw_data.columns[2]}")
```
ireneisdoomed, line 276: Is this correct?
```suggestion
            raise ValueError(f"Unrecognized label separator in {raw_data.columns[3]}")
```

partner_preview_scripts/ot_crispr.py
```diff
@@ -241,7 +276,7 @@ def _read_and_filter_mageck_file(
...
 
         # Updating column names according to the identified label separator:
-        raw_data = reduce(
+        return reduce(
             # Rename all columns:
```
ireneisdoomed, line 280: Is it only `pos|p-value` that needs to be ammended?

partner_preview_scripts/ot_crispr.py
```diff
@@ -226,13 +211,63 @@ def _read_and_filter_mageck_file(
         Raises:
             ValueError: If the label separator is not recognized.
         """
-        side = filter_column.split("|")[0]
+        # Reading input data and immediately fix the headers:
+        raw_data = self.adjust_column_names(
```
ireneisdoomed, line 215: If we only need to rename one column, having to pass the entire dataframe to do a column renaming looks very verbose.
Could we do sth like:
```
raw_data = self.spark.read.csv(mageck_file, header=True, sep="\t")

p_value_col = [col for col in raw_data.columns if col.startswith("pos")][0]
p_value_col_norm = p_value_col if "|" in p_value_col else p_value_col.replace(".", "|")

raw_data = (
raw_data
.withColumnRenamed(p_value_col, p_value_col_norm)
...
)
``` 

If it's more than one, something like this looks simpler too and it is more readable (at the expense of not controlling any other possible separators):
```
def normalize_column_names(df: DataFrame) -> DataFrame:
    for col_name in df.columns:
        if "." in col_name:
            df = df.withColumnRenamed(col_name, col_name.replace(".", "|"))
    return df
```

partner_preview_scripts/ot_crispr.py
```diff
@@ -14,6 +14,7 @@
 import logging
 from functools import reduce
 from typing import Optional
+import operator
```
ireneisdoomed, line 17: I don't think this is used?

ireneisdoomed (dismissed):
Thank you for solving this. I've seen it works.
I have 2 questions, though:
- Can you give me an example of a value in filterColumn? I don't really understand how this is handled
- "At the next PPP ETL run, the updated dataset will be picked up." Should we point to a different input in the configuration?

partner_preview_scripts/ot_crispr.py
DSuveges: I wanted to make sure the process fails at this point if an unexpected separator is found. I am not checking all the mageck input before running. There's a risk that the files we get might diverge. I was also considering some sort of pre-flight data validation, but the value is not big enough.

ireneisdoomed:
The parser migration is based on the code from this branch. I suggest merging

ireneisdoomed (approved):

